PR #7, opened 2016-01-26
Stop using spans for word wrap

stianjensen:
Using css instead results in less html markup, and more readable source
code.


lionleaf:
Conflicts!

stianjensen:
ping

PR #31, opened 2016-05-19
Don't start demo if it is paused

iver56:
... so that demos that are out of view get the first loop call when they are actually scrolled into view


lionleaf:
This leads to a blinking effect when scrolling through though. In most cases for visual demos it's better to have the first frame already rendered, no?

I get that it's less than optimal for your prompt demo, but using prompt is not really the purpose of this.

Thoughts?

PR #39, opened 2016-07-18
Change time calculation to avoid rounding errors when calc frame

lionleaf:


iver56:
Looks like this broke the fix introduced here:
https://github.com/lionleaf/dwitter/pull/27

This snek isn't very snek when it doesn't move ;)
http://dweet.dwitter.net/id/197?autoplay=1

lionleaf:
Oh, right ;P Nice catch!

PR #42, opened 2016-07-19
Move delete to POST with csrf

lionleaf:


dwitter/feed/views.py
```diff
@@ -81,7 +81,8 @@ def dweet_reply(request, dweet_id):
 
 
 @login_required
-def dweet_delete(request, dweet_id):
+def dweet_delete(request):
+    dweet_id = request.POST['dweet-id']
```
sigvef: Ideally you could use django forms to parse POST data in order to get free systematic validation of input.

dwitter/feed/views.py
```diff
@@ -81,7 +81,8 @@ def dweet_reply(request, dweet_id):
 
 
 @login_required
-def dweet_delete(request, dweet_id):
+def dweet_delete(request):
```
sigvef: Consider specifying that this view is post-only for safety and completeness.

PR #43, opened 2016-07-20
Use Fullscreen API for fullscreen button

### LINKED ISSUE #38: Fullscreen the canvas directly
Using fullscreen APIs to fullscreen the actual canvas instead of opening a new tab

sigvef:
Non-js clients will gracefully fall back to regular old linking. A
special-case is needed in the scrolling play/pause logic, since
Waypoints.inview thinks that fullscreened elements are not inview.

This fixes #38.

### DIFF
--- a/dwitter/static/js/feed.js
+++ b/dwitter/static/js/feed.js
@@ -7,4 +7,19 @@ var onDweetChanged = function() {
 
 $(document).ready(function()  {
   $('body').on('input','.code-input', onDweetChanged);
+
+
+  function requestFullscreen(el) {
+    (el.mozRequestFullScreen ||
+     el.webkitRequestFullscreen ||
+     el.requestFullscreen).call(el);
+  }
+  $('.dweet').each(function(i, el) {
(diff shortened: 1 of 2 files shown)


iver56:
> Non-js clients will gracefully fall back to regular old linking

Also, non-js clients won't be able to show any of the dweets in the first place :stuck_out_tongue_winking_eye:

dwitter/static/js/feed.js
```diff
@@ -7,4 +7,19 @@ var onDweetChanged = function() {
...
 $(document).ready(function()  {
   $('body').on('input','.code-input', onDweetChanged);
+
+
+  function requestFullscreen(el) {
+    (el.mozRequestFullScreen ||
+     el.webkitRequestFullscreen ||
+     el.requestFullscreen).call(el);
+  }
+  $('.dweet').each(function(i, el) {
```
iver56: This needs to be done again for new dweets loaded with AJAX

PR #45, opened 2016-07-20
Add missing urls for password management.

### LINKED ISSUE #22: Reset Password not working
Attempting to reset you password results in a Server Error (500)

lionleaf:
Fixes #22


stianjensen:
This seems weird. Doesn't django-registration provide all password views?

sigvef:
Yes and no. I ended up doing something like this in ideavote.co, fwiw:

image

PR #60, opened 2016-08-18
Add dbbackup

lionleaf:


requirements.txt
```diff
@@ -1,8 +1,9 @@
-django==1.8
+django<1.9
```
sigvef: It is a really good idea to only specify exact version numbers here. That way, you won't run into issues where differing versions in prod and dev create subtle and hard-to-discover bugs due to differences in versions. You can use pip freeze to see the current version numbers installed in an environment.

lionleaf: Hmm... Good point, thanks. How do you deal with keeping stuff up to date? Wait till you need it, or run regular upgrades?
This was triggered by a bug in 1.8.0 (I meant to use the newest 1.8 when writing it initially, but 1.8 means 1.8.0)

stianjensen: Haven't tried this, but you could check out https://doppins.com

PR #64, opened 2016-08-19
Implement reddit hot algorithm (and add timestamps to dweets)

### LINKED ISSUE #62: More advanced sorting algorithm for "hot" dweets
Something like the [reddit ranking algorithm](https://medium.com/hacking-and-gonzo/how-reddit-ranking-algorithms-work-ef111e33d0d9#.n30qlo7sg) or  [hackernews ranking algorithm](https://medium.com/hacking-and-gonzo/how-hacker-news-ranking-algorithm-works-1d9b0cf2c08d#.hzrvyvsi6).

Currently leaning towards the reddit one, as the values doesn't decrease over time, meaning you only have to recalculate the score for a single post every time it gets a like.

lionleaf:
fixes #62 

Implements reddits hot algorithm as of https://medium.com/hacking-and-gonzo/how-reddit-ranking-algorithms-work-ef111e33d0d9#.n30qlo7sg .

Numbers changed a bit so that it will take 24 * log_2(number of "awesome") hours before a post is being ranked under a post with 0 "awesome". Essentially the post gets 24*log_2(awesome) extra hours added to its timestamp and then you sort by this modified timestamp.

I'm very open to feedback on these numbers. They're set to work a bit slower than reddit (which is 12 * log_10(upvotes - downvotes) ), because we have both fewer votes and fewer posts.

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -9,8 +9,19 @@ from dwitter.models import Dweet
 from django.contrib.auth.decorators import login_required
 from django.utils import timezone
 from functools import wraps
+from datetime import datetime, timedelta
+from math import log
 import json
 
+def epoch_seconds(date):
+    epoch = datetime(2015,5,5)
+    naive = date.replace(tzinfo=None)
+    td = naive - epoch
+    return td.days * 86400 + td.seconds + (float(td.microseconds) / 1000000)
+
+def hot(likes, date):
+    order = log(max(abs(likes), 0), 2)
+    return round(order + epoch_seconds(date)/86400)
 
 def ajax_login_required(view_func):
     @wraps(view_func)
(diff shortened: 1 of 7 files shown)


dwitter/feed/views.py
```diff
@@ -9,8 +9,19 @@
 from django.contrib.auth.decorators import login_required
 from django.utils import timezone
 from functools import wraps
+from datetime import datetime, timedelta
+from math import log
 import json
 
+def epoch_seconds(date):
+    epoch = datetime(2015,5,5)
```
sigvef: Why did you choose this date for epoch?

lionleaf: Just an arbitrary date before dwitter.

I guess I could've gone for 2000,1,1 or something more memorable.

Issue #78, opened 2016-10-01
First comment by author as a sticky comment on top


lionleaf:
When posting a dweet, allow a description that is posted as the first comment. This comment should then be "sticky" and emphasized.

iver56:
Good idea, but be careful. Before a draft is restored, ask the user if he/she wants to apply it. The code in the draft may be heavy or bad code that was the reason why the user shut down the browser tab.

lionleaf:
This was probably a response to #92

PR #91, opened 2017-01-14
Make first comment sticky if it's by the dweet author

lionleaf:
Instagram-style. Allows users to post a comment about the
dweet that will be highlighted and always visible

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,4 +17,5 @@ Current build: [dwitter.net](http://dwitter.net)
 * `make migrate`
 * `make lint`
 * `make shell`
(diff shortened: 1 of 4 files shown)


lionleaf:
fixes #78

sigvef:
🚢 LGTM (don't have permission to merge myself)

iver56:
Looks like the first comment is shown twice, with the first one in bold

lionleaf:
Yup. Only for when there is between 1 and 3 comments, but I'm on it.

Issue #95, opened 2017-02-20
Inject more Math functions


naughton:
Would be more interesting if we had more Math functions injected...

    var A = Math.abs;
    var S = Math.sqrt;
    var G = Math.sign;
    var r = Math.random;

and maybe even:

    var d = x.drawRect

and

    var F = (s) => x.fillStyle = s;

bakkot:
Seconding `Math.random` and `fillRect` especially.

naughton:
implemented this in PR #98 
A = Math.abs
G = Math.sign
Q = Math.sqrt;
D = (scale=1) => Math.random() * scale  // D for "dice"?
B = (x,y,w,h) => x.fillRect(x,y,w,h)            // B for "box"?
P = (x,y) => x.fillRect(x,y,1,1)                     // P for "point"?
F = (s) => x.fillStyle = s

lionleaf:
Thanks a lot for the pull request! This has been discussed a few times, and it is quite tempting, but I'm a bit hesitant as it will diverge even more from normal js and it's effectively the same as increasing the character count. To be fair the "purity" is already broken by S, C etc. So I'm open to a discussion

sigvef:
One of the ideas in the beginning was that it would force you to examine the docs to find seldom-used properties to exploit, so that it also doubled as a learning tool for the APIs.

johanberonius:
And in that case, maybe more math constants with single greek letter variable names, e.g.:
`π = Math.PI`
`τ = Math.PI * 2`

sigvef:
Using `7` instead of `Math.PI * 2` often works suite well :P

donbright:
At first, I had many functions I would also like added. x*x+y*y, x*x-y*y, 2*x*y, etc. 

But after consideration, I am against adding anything. The beauty is in the simplicity. Let it be simple. If people want complexity there is Processing and WebGL and three.js and many other find and wonderful complex systems.  

[...]

PR #98, opened 2017-02-20
added more functions from math and ctx

naughton:
see #95


lionleaf:
Thanks!

I have however decided to close this pull request for now. See #95 for discussion.

Issue #12, opened 2016-02-21
Make infinite scroll more obvious


lionleaf:
Should be a "loading ... " when at the bottom to indicate that more items are about to load.

Also consider a fall-back link to manually load more

lionleaf:
Resolved by #101

PR #101, opened 2017-02-22
Feed loading indicator

tbroadley:
This PR is in response to https://github.com/lionleaf/dwitter/issues/12. It adds a loading indicator to the bottom of feeds when loading new posts. Also, when a user reaches the end of a feed, a message is displayed.

Demo:

dwitter-infinite-scroll-loading

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -83,6 +83,7 @@ def feed(request, page_nr, sort):
                'header_title': 'Dwitter',
                'feed_type': 'all',
                'page_nr': page,
+               'on_last_page': last == dweet_count,
                'next_url': next_url,
                'prev_url': prev_url,
                'sort': sort,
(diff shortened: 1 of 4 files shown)


dwitter/feed/views.py
```diff
@@ -83,6 +83,7 @@ def feed(request, page_nr, sort):
                'header_title': 'Dwitter',
                'feed_type': 'all',
                'page_nr': page,
+               'on_last_page': last == dweet_count,
```
lionleaf: Should be 'last >= dweet_count' as last is simply "page * dweets_per_page" so I'm guessing this will fail when the number of dweets are not a multiplum of 10.

tbroadley: [Lines 61 and 62](https://github.com/tbroadley/dwitter/blob/b489a4fa509f74fd0c2638a5647acbccd0d174bf/dwitter/feed/views.py#L61-L62):

```py
if (last >= dweet_count):
    last = dweet_count
```

So if `last` was 20 and `dweet_count` was 15, `last` will be set to 15. Then we check if `last == dweet_count`, which is true.

lionleaf:
Oh, good point. I read it a bit too quick. I'll merge when I get home from
work today. Thanks for the PR

tbroadley:
No problem! Thank you for the quick review.

Issue #106, opened 2017-02-22
Improve 404 page


iver56:
Currently it says
```
Not Found

The requested URL /d/904 was not found on this server.
```

It would be better if it showed a more user-friendly message, and suggested a link to the front page or something. Could also apply the same style (css) as on other dwitter pages.

Bonus: Maybe show a funny 404-related dweet or something?

lionleaf:
Solved by #112 . Created #116 for your bonus idea 👍

PR #112, opened 2017-02-24
404 page

gnocchi2815:
I've created a basic but improved 404 page, in response to issue #106. Perhaps someone would like to improve on this further by adding a 404 related tweet as suggested in the issue.

### DIFF
--- a/dwitter/templates/404.html
+++ b/dwitter/templates/404.html
@@ -1,0 +1,7 @@
+{% extends 'base.html' %}
+
+{% block content %}
+
+    <h1>This dweet doesn't exist</h1>
+
+{% endblock %}
--- a/dwitter/templates/404.html
+++ b/dwitter/templates/404.html
@@ -2,6 +2,6 @@
 
 {% block content %}
 
-    <h2 class=header-title>This dweet doesn't exist.</h2>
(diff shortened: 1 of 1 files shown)


dwitter/templates/404.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'base.html' %}
+
+{% block content %}
+
+    <h2 class=header-title>This dweet doesn't exist.</h2>
```
lionleaf: You'd get a 404 for other cases than missing dweets. Maybe change to "this page doesn't exist"

gnocchi2815: Thanks for the feedback - good point!

gnocchi2815: Updated

lionleaf:
Thanks! Deployed: https://www.dwitter.net/lalalalakwd :)

lionleaf:
Currently had to disable it for a while. It was used for dweet.dwitter.net as well (where I host the dweet content) and someone wrote a dweet that references a random one. Because dweet.dwitter.net doesn't have the same auth session it would crash and flood my inbox. Will restore the 404 once I have time to fix that

Issue #108, opened 2017-02-24
Resize textbox for dweets with many lines


lionleaf:
After the change to using space instead of ; in the example dweet, a lot of people make dweets with many lines. Textarea should resize to fit this (probably not all the way to 140 newlines...)

PR #114, opened 2017-02-24
Add auto-resize to textareas

EpicOrange:
Addresses #108.


lionleaf:
Thanks again!
Seems to work well on Chrome when editing :) Two comments:

1. The size is wrong when a multiline dweet is loaded in the feed (it never invokes 'oninput' as the dweet code is embedded in the django template)

2. This allows very tall dweets, up to 140 characters, which would be multiple pages long. While such dweets probably won't climb high, they can be a bit disruptive in the 'new' feed and a maximum textarea height would be a good idea. I'd suggest 10 lines max.

Let me know if you want to take a stab at those points as well. I'd be happy to help you out with those, but it probably won't be until sometime next week


screenshot 2017-02-25 03 46 17

EpicOrange:
Glad it works well, but these two issues cancel themselves out as intended. Even if you post a 140-line long tweet, they won't be disruptive to viewers until they actually go in and edit them!

True, it's a bit excessive to have 140 lines maximum. 10 lines is a good maximum for finished dweets; I can't imagine many dweets will take up more than 10 lines. But while I'm editing, I like to have space, so I'd increase that to 20 lines (which shouldn't be too disruptive).

Issue #109, opened 2017-02-24
Soft-delete instead of hard-delete dweets


lionleaf:
Instead of deleting from the database, leave a shell of the dweet with the dweet flag. This way, "remix of" chains aren't broken and direct links still work (showing "this dweet is deleted"), while it's still removed from the feed.

stianjensen:
This was completed in #120, I believe.

PR #120, opened 2017-02-26
Soft delete for dweets

tbroadley:
In response to https://github.com/lionleaf/dwitter/issues/109.

This PR introduces soft delete for dweets. When a dweet is deleted, instead of being removed from the database, its `deleted` value is set to `True`. Deleted dweets are not shown in feeds. When accessed through a `/d/[id]` URL, a special page is shown, with just the date and time posted and a link to the parent dweet (if any). Also, liking, remixing, deleting, or accessing the fullscreen view of a deleted dweet results in a 404.

~I think it's a little ugly to call `.filter(deleted=False)` all over the place. If there is a way that results in less code duplication, let me know!~ Fixed with a custom `Manager` class.

~Even with this change, if one dweet in a "remix-of" chain is deleted, it's impossible to follow the chain all the way to the root dweet. I'm not sure if this is a good behaviour or not, to be honest. Maybe the 404 page for the deleted dweet could link to that dweet's parent.~ Fixed by adding a deleted dweet page.

It would also make sense to soft-delete dweets when the author's account is deleted. Right now, we still do a permanent delete in that case. ~I'd prefer to make that change in a separate PR.~ I'll make this change in this PR.

## Todo:

- [x] Soft-delete dweets when the author's account is deleted

## Demo:

dwitter-soft-delete-2

image

image


sigvef:
> I think it's a little ugly to call .filter(deleted=False) all over the place. If there is a way that results in less code duplication, let me know!

You can use Django's Managers for something like this: https://docs.djangoproject.com/en/1.10/topics/db/managers/#modifying-a-manager-s-initial-queryset

sigvef:
> Deleted dweets are not shown in feeds, and cannot be accessed directly through a /d/[id] URL.

Perhaps they should be accessible via the direct url, but the content and author of the dweet is redacted, like with reddit comments:

image

dwitter/models.py
```diff
@@ -1,16 +1,27 @@
...
-    reply_to = models.ForeignKey("self", on_delete=models.SET_NULL,
+    reply_to = models.ForeignKey("self", on_delete=models.DO_NOTHING,
                                  null=True, blank=True)
 
     author = models.ForeignKey(User, on_delete=models.CASCADE)
     likes = models.ManyToManyField(User, related_name="liked")
     hotness = models.FloatField(default=1.0)
+    deleted = models.BooleanField(default=False)
+
+    objects = NotDeletedDweetManager()
```
tbroadley: This changes the default `Dweet.objects` manager to filter out deleted dweets.

dwitter/migrations/0013_auto_20170226_1249.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('dwitter', '0012_auto_20170225_2056'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
lionleaf: Applying the migration unlinks all dweets:

screenshot 2017-02-28 08 50 53

lionleaf: Removes author and creates a new field _author that is just set to null. Posting new dweets work though, and the soft delete seems to work :)

tbroadley: Oops, good catch! I'll fix the problem with existing dweets being unlinked from their authors.

tbroadley:
OK, I think this PR is ready to go. I tested the migration, and dweets seemed to stay associated with their authors.

lionleaf:
Thanks!

The migration seems to have gone smoother now. Successfully migrated as well as rolling back the migration.

However, I'm getting 500 errors when trying to post comments on a dweet. Also, have you considered the behaviour of comments upon dweet / user deletion?

tbroadley:
I'll look into the 500 errors when posting comments. Thanks for checking that! I'll make sure to manually test out everything before updating this PR again. 

Currently, comments are hard-deleted when their dweet or user is deleted. I think this is OK, since we don't display comments on deleted dweets, and dweets are soft-deleted when their creator is deleted.

dwitter/models.py
```diff
@@ -24,7 +62,15 @@ class Comment(models.Model):
...
-    author = models.ForeignKey(User, on_delete=models.CASCADE)
+    _author = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    @property
+    def author(self):
+        return self._author
+
+    @author.setter
+    def author(self, value):
+        self._author = value
```
tbroadley: The 500 error was caused by an error with the following message:

```
"_author" is an invalid keyword argument for this function.
```

I couldn't determine exactly why this was happening. Changing `Comment` to use `_author` instead of `author` seems to fix the problem.

PR #124, opened 2017-02-26
Add user avatars from gravatar

sigvef:
image

image

image

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -4,6 +4,18 @@
   padding:0px;
 }
 
+.avatar {
+  width: 36px;
+  height: 36px;
+  border-radius: 18px;
+  background: #ccc no-repeat center;
+  background-size: cover;
+  overflow: hidden;
+  display: inline-block;
+  vertical-align: middle;
+  box-shadow: 0 0 2px rgba(0, 0, 0, 0.2);
+}
+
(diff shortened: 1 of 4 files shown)


tbroadley:
On small screens, the dropdown menu is rendered behind the blank dweet's `iframe`.

image

tbroadley:
Also, when logged out, the "Login or register" links are still styled with a darker background when hovered over:

image

tbroadley:
Rather than displaying the Gravatar logo, maybe the Gravatar image should be hidden when there is none associated with the user's email address.

sigvef:
> On small screens, the dropdown menu is rendered behind the blank dweet's iframe.

Fixed.

> Also, when logged out, the "Login or register" links are still styled with a darker background when hovered over:

Fixed.

> Rather than displaying the Gravatar logo, maybe the Gravatar image should be hidden when there is none associated with the user's email address.

Now the default is Gravatar's auto-generated retro 8-bit images (based off of your email, so different users get different avatars):

image

tbroadley:
Looks sweet! One more comment: On small screens, the current user's name in the top right (which is also a link to the user's profile) is hidden. I think that the Gravatar image should also link to the user's profile, so that it is still possible to access one's profile.

PR #126, opened 2017-02-26
only allow POST to like endpoint; add CSRF token

tbroadley:
This PR changes endpoints for liking dweets so that only POST requests are accepted. This prevents dweets like https://www.dwitter.net/d/1058 from causing users to inadvertently like/unlike dweets.

I've also added a CSRF token to dweet-liking requests.

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -154,6 +154,7 @@ def dweet_delete(request, dweet_id):
 
 
 @ajax_login_required
+@require_POST
 def like(request, dweet_id):
     dweet = get_object_or_404(Dweet, id=dweet_id)
 
--- a/dwitter/static/js/ajax-handling.js
+++ b/dwitter/static/js/ajax-handling.js
@@ -1,4 +1,17 @@
-
+function getCookie(name) {
+  if (document.cookie && document.cookie !== '') {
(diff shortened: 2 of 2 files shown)


sigvef:
Looks good!

lionleaf:
Does the CSRF protection work? I was under the impression you need @csrf_protect ?

PS: postComment in ajax-handling.js does gets the CSRF value in a different way, basically embedding with the django {% csrf_token %} template magic

tbroadley:
Apparently it's not necessary to decorate with `csrf_protect` if the CSRF middleware is included in the list of middlewares.

I'll take a look at how `postComment` is passing the CSRF token, and change `processLike` to handle it that way. Thanks for letting me know that there's a better way to get the token than from `document.cookie`!

sigvef:
The Django docs recommend using the cookie approach, interestingly: https://docs.djangoproject.com/en/1.10/ref/csrf/#ajax

tbroadley:
True! If there were more places where AJAX requests were made, then it would be nice to automatically add the CSRF token to all AJAX requests. As it is, there are only two places were AJAX POST requests are made. Right now, I don't think it particularly matters which way it's done. Definitely something to revisit if more AJAX POST requests are added, IMO.

Issue #125, opened 2017-02-26
Format dweet posted times in the user's time zone


tbroadley:
If users had a way to set their timezone, dweet post times could be displayed in their time zone.

iver56:
This can also be auto-detected, right? I guess that would be more convenient

tbroadley:
That's true, auto-detection would be better. AFAIK it's quite difficult to do auto-detection on the backend. One solution would be to run some Javascript on page load that would replace all times on the page with the corresponding times in the user's timezone.

sigvef:
https://momentjs.com/ is king of the hill for this kind of stuff ^^

tbroadley:
❤️ moment.js!

PR #128, opened 2017-02-26
use moment.js to format dweet timestamps in user's timezone

tbroadley:
In response to https://github.com/lionleaf/dwitter/issues/125.

This PR also changes the date format slightly: from `Feb. 26, 2017, 7:36 p.m.` to `Feb 26, 2017 at 7:36 PM UTC`. If JS is disabled, the date and time will still be displayed in UTC.

image


iver56:
How about using localized format (format string `lll`) instead?
https://momentjs.com/docs/#localized-formats
That way, norwegians (like me) get 24-hour clock without AM/PM, for example

tbroadley:
Here's what localizing the posted time (for the Norwegian `nb-NO` locale) looks like:

image

I noticed that Moment renders the date in the language specified with the locale. This is inconsistent with the rest of Dwitter, which is in English. We could use a non-localized format for te date, then use the `LT` time format (which is localized).

lionleaf:
I forgot the test that always breaks things; scrolling down and dynamically loading dweets.

If you load a new page of dweets, the timezone js isn't invoked and it is now displayed in UTC. There is some code duplication for dynamically loaded dweets that I should get rid of in scrolling.js. 

I'll get back to this after work tomorrow (which is in UTC time ;) ).

tbroadley:
Shoot, I should have thought of that. Sorry @lionleaf.

Issue #129, opened 2017-02-26
Tests


tbroadley:
Let's work together and add some tests to Dwitter! :smile:

PR #130, opened 2017-02-26
basic tests for the Dweet model

tbroadley:
In response to https://github.com/lionleaf/dwitter/issues/129.

### DIFF
--- a/dwitter/tests/test_dweet.py
+++ b/dwitter/tests/test_dweet.py
@@ -39,9 +39,9 @@ class DweetTestCase(TestCase):
         Dweet.objects.get(id=2)
 
     def test_dweet_has_correct_likes(self):
-        self.assertEqual(Dweet.objects.get(id=1).likes.count(), 0)
-
+        dweet1 = Dweet.objects.get(id=1)
         dweet2 = Dweet.objects.get(id=2)
-        self.assertEqual(dweet2.likes.count(), 2)
-        self.assertIn(User.objects.get(username="user1"), dweet2.likes.all())
-        self.assertIn(User.objects.get(username="user2"), dweet2.likes.all())
+        all_users = [repr(u) for u in User.objects.all()]
+
+        self.assertQuerysetEqual(dweet1.likes.all(), [])
+        self.assertQuerysetEqual(dweet2.likes.order_by('id'), all_users)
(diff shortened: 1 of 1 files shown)


dwitter/tests/test_dweet.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def test_dweet_reply_to_set_null_on_delete(self):
+        Dweet.objects.get(id=1).delete()
+        self.assertEqual(Dweet.objects.get(id=2).reply_to, None)
+
+    def test_dweet_author_cascade_on_delete(self):
+        User.objects.get(username="user1").delete()
+        with self.assertRaises(Dweet.DoesNotExist):
+            Dweet.objects.get(id=1)
+        Dweet.objects.get(id=2)
```
tbroadley: If the dweet with `id=2` were deleted, this line would throw an exception.

dwitter/tests/test_dweet.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def test_dweet_has_correct_likes(self):
+        self.assertEqual(Dweet.objects.get(id=1).likes.count(), 0)
+
+        dweet2 = Dweet.objects.get(id=2)
+        self.assertEqual(dweet2.likes.count(), 2)
```
gnocchi2815: This would be neater if both lines were put into the same format e.g. 
`self.assertEqual(Dweet.objects.get(id=1).likes.count(), 0)`
`self.assertEqual(Dweet.objects.get(id=2).likes.count(), 2)`

dwitter/tests/test_dweet.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertIn(User.objects.get(username="user1"), dweet2.likes.all())
+        self.assertIn(User.objects.get(username="user2"), dweet2.likes.all())
```
gnocchi2815: I think this should also test that a dweet has been liked only once by each user.

tbroadley: I agree. The test doesn't do a great job of expressing that it tests whether each user has liked the dweet. I'll try to improve that!

dwitter/tests/test_dweet.py
tbroadley: It would be neater, but at the cost of duplicating `Dweet.objects.get(id=2)`. I don't think it's worth it.

lionleaf: You could just do:
dweet1 = [...]
To get the symmetry without duplicating Dweet.objects.get(id=2)

PR #138, opened 2017-02-28
Display errors when evaluating or running dweets

EpicOrange:
Now you don't have to open dev console to see errors! (Addresses #77.)

The error div is contained inside the dweet iframe, so the iframe now has both a canvas and a div. Dweet code can potentially select and modify the div but this shouldn't be harmful.

Demo:
demo

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -10,6 +10,7 @@
       color: red;
       font-size: small;
       position: absolute;
+      font-family: sans-serif;
       top: 1em;
       left: 1em;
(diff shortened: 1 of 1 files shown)


tbroadley:
Nice! I'd suggest making the error message's font the same as the font used in the rest of the app.

sigvef:
Cool stuff! I just ran into this dweet which has apparently relied on secretly crashing a little, and now it has a big red error! I wonder how many old dweets are affected by this ^^

image

EpicOrange:
Thanks! Addressed this in #159.

PR #143, opened 2017-02-28
Reduce user menu hover target

sigvef:
Before this commit, you could hover over the invisible parts of the drop
down menu to make it visible. That was a problem because the invisible
part would cover the "hot/new/top" links on small screens, making them
unclickable. Now, only the username/avatar can be hovered over to
trigger the menu.

Before:

After:

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -274,13 +274,14 @@ iframe {
 }
 
 .head-account-info .collapsible {
-  opacity: 0;
   position: absolute;
   border: 1px solid #ddd;
   right: 1em;
-  margin-top: -4px;
-  transition: all 0.05s;
   z-index: 10;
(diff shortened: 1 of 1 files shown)


dwitter/static/main.css
```diff
@@ -274,13 +274,11 @@ iframe {
 }
 
 .head-account-info .collapsible {
-  opacity: 0;
   position: absolute;
   border: 1px solid #ddd;
   right: 1em;
-  margin-top: -4px;
-  transition: all 0.05s;
```
tbroadley: I think it would be nice to keep this transition. Would changing it to `transition: opacity 0.05s` work?

sigvef: I had to be a little sneaky with transitions to get the opacity back in (since you can't transition display none/block), but it worked \o/

PR #144, opened 2017-02-28
User settings

sigvef:


peterhgombos:
Do we need full names? That's not very demosceney, right?

lionleaf:
Optional full names isn't a problem, but then again there isn't a need. I suggest we just do the email address and the avatar preview/instruction for now. We can add names if we ever decide to show them somewhere :)

sigvef:
I agree, I just added them since they were on the model anyway. Will remove.

sigvef:
Removed.

PR #148, opened 2017-02-28
Make code area dark, tweak dweet card design

sigvef:
This tightens up the bottom part of the dweet card design. In the future
we can add syntax hilighting to the dweet code!

(I'd have a screenshot here, but github file uploads don't work for me right now)


sigvef:
rebased

image

lionleaf:
Oh, didn't notice you moved the "delete" button at first.

lionleaf:
Suggestion: Can you make the dark area look like it's from a deeper layer behind the dweet card?

Oh, you merged while I was typing this ^^

PR #169, opened 2017-03-06
rename _author back to author

tbroadley:
On its own, this PR could lead to 500 errors occurring. Dweets with an author whose account was deleted before the migration would now have a `NULL` author, with no defaulting to `get_sentinel_user`. To fix this, I'd suggest running the following script before doing the migration:

```py
import os

os.environ.setdefault("DJANGO_SETTINGS_MODULE", "dwitter.settings")

import django
from dwitter.models import Dweet
from dwitter.models import get_sentinel_user

django.setup()

for dweet in Dweet.with_deleted.filter(author=None):
    author = get_sentinel_user()
```

### DIFF
--- a/dwitter/migrations/0013_auto_20170305_1830.py
+++ b/dwitter/migrations/0013_auto_20170305_1830.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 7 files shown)


tbroadley:
Merge conflict resolved :)

Issue #16, opened 2016-02-21
fps-counter in edit mode


lionleaf:

stianjensen:
This tool might be useful: https://github.com/mrdoob/stats.js/

PR #174, opened 2017-03-07
add an FPS counter to new and edited dweets

tbroadley:
This resolves https://github.com/lionleaf/dwitter/issues/16.

Demo:

fps-counter


sigvef:
Perhaps we should have a toggle for it as well, in case it covers up a part of the demo that you want to see while editing? We could add that later though.

tbroadley:
I'll create an issue for adding a way to disable and enable the FPS counter :+1: it's definitely a good thing to have.

PR #186, opened 2017-03-12
Redesign and cleanup header

stianjensen:
This reduces the height of the header by 50px, making the content of the
page more prominent.
It also reduces the font size of the dropdown menu so every item in it
can fit on one line.

screenshot 2017-03-12 18 53 41

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -247,6 +247,7 @@ iframe {
   width:100%;
   height:100%;
 }
+
 .head-menu {
   width:100%;
   background:white;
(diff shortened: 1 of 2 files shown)


stianjensen:
> Let's max-width the header, so it doesn't get all out of whack on larger screens.

Done

lionleaf:
Seems to break the drop-down menu when I tried it:

screenshot 2017-03-12 19 19 28

stianjensen:
> Seems to break the drop-down menu when I tried it:

Yup, just caught that myself (it was due to the max-widh). Should work again now

lionleaf:
Nice!

Still some dropdown issues:
screenshot 2017-03-12 19 24 35

stianjensen:
> Still some dropdown issues:

Lol, of course. Now the fix became a bit more complicated. I'll rethink.

stianjensen:
ping.

Actually, the fix was easy. overflow hidden was not necessary as a clearfix hack, since the height of the header is specified to be 40px anyway.

Issue #132, opened 2017-02-27
Local timezones doesn't work for dynamically loaded dweets.


lionleaf:
#128 introduced timestamps in local time. However, this doesn't work at the moment for dweets that are loaded after scrolling to the bottom. (Currently live)

PR #188, opened 2017-03-12
format times in local timezone on loaded dweets

tbroadley:
This closes https://github.com/lionleaf/dwitter/issues/132.

### DIFF
--- a/dwitter/static/js/scrolling.js
+++ b/dwitter/static/js/scrolling.js
@@ -43,6 +43,11 @@ window.onload = function() {
               $(this).off("click.a keyup.a").select();
             });
           });
+
+          moment.locale(navigator.userLanguage || navigator.language || 'en-US');
+          var timestamp = $(div).find('time');
+          var postedDate = moment.utc(timestamp.attr('datetime'));
+          timestamp.text(postedDate.local().format('lll'));
         });
       }
   });
--- a/dwitter/templates/snippets/deleted_dweet_card.html
+++ b/dwitter/templates/snippets/deleted_dweet_card.html
@@ -9,6 +9,10 @@
         {% include 'snippets/remix_info.html' %}
       {% endif %}
     </div>
-    <p class=dweet-timestamp> {{ dweet.posted }} </p>
+    <p class="dweet-timestamp">
+      <time datetime="{{ dweet.posted | date:"Y-m-d H:i" }}">
+        {{ dweet.posted | date:"M j Y g:i A" }} UTC
+      </time>
+    </p>
   </div>
 </div>
--- a/dwitter/static/js/feed.js
+++ b/dwitter/static/js/feed.js
@@ -54,7 +54,7 @@ $(document).ready(function()  {
 
   moment.locale(navigator.userLanguage || navigator.language || 'en-US');
   $('.dweet-timestamp').each(function(_, element) {
-    var postedDate = moment.utc($(element).text(), 'MMM D YYYY h:mm A [UTC]');
+    var postedDate = moment.utc($(element).find('time').attr('datetime'));
     $(element).text(postedDate.local().format('lll'));
   });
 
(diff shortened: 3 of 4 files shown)


dwitter/static/js/scrolling.js
```diff
@@ -43,6 +43,11 @@ window.onload = function() {
               $(this).off("click.a keyup.a").select();
             });
           });
+
+          moment.locale(navigator.userLanguage || navigator.language || 'en-US');
+          var timestamp = $(div).find('.dweet-timestamp');
+          var postedDate = moment.utc(timestamp.text());
+          timestamp.text(postedDate.local().format('lll'));
```
tbroadley: This code is mostly duplicated from `feed.js`. At some point, I think that file and `scrolling.js` should be refactored. Since we don't have any tests covering those files, though, I'm leery of doing that refactor right now.

dwitter/templates/snippets/deleted_dweet_card.html
```diff
@@ -9,6 +9,6 @@
         {% include 'snippets/remix_info.html' %}
       {% endif %}
     </div>
-    <p class=dweet-timestamp> {{ dweet.posted }} </p>
+    <p class=dweet-timestamp>{{ dweet.posted | date:"Y-m-d H:i" }}</p>
```
tbroadley: While working on this I ran into what I think is a bug with moment. The library doesn't seem to apply daylight savings time properly to dates parsed with an explicit format string:

```
moment.utc("Mar 12 2017 3:33 PM", "MMM D YYYY h:mm A").local().format()
> "2017-01-12T10:33:00-05:00"

moment.utc("2017-03-12 15:33").local().format()
> "2017-03-12T11:33:00-04:00"
```

That's why I switched the date format to this ISO 8601 date format, which is recognized automatically by moment.

tbroadley: Lucky that we sprang forward today! I wouldn't have noticed this issue otherwise.

stianjensen:
For consistent parsing with moment, it's recommended to always input ISO 8601 formatted strings (like 2016-03-12T20:46:00Z).
You can use date:"c" in Django to output this format.
This strings aren't usually very pretty, so one thing you can do is output things from Django in this format:
```
<time datetime="{{ dweet.posted | date:"c" }}">{{ dweet.posted | date:"M j Y g:i A" }}</time>
```

tbroadley:
Nice, I didn't know about the `time` HTML element. Thanks @stianjensen!

Unfortunately, moment doesn't automatically recognize Django's ISO 8601 format. That's why I didn't use the `c` option to format the date.

stianjensen:
> Unfortunately, moment doesn't automatically recognize Django's ISO 8601 format. That's why I didn't use the c option to format the date.

That's a bit surprising, since ISO 8601 is _the_ standard date format. Could it be because we're using `moment.utc(date)` instead of `moment(date)`?

stianjensen:
For reference, here's how I've implemented moment js in a django app before: https://github.com/stianjensen/wikipendium.no/pull/370/files

tbroadley:
No, I don't think it's because we are using `moment.utc`. I believe it's because Django's format for seconds has six digits after the decimal point, whereas moment only expects three.

In my browser, moment does parse the Django date format correct, but throws a warning indicating that it's falling back to calling `new Date` on the string. Since the behaviour of `new Date` varies from browser to browser, I'd rather give moment the date in a format that it can definitely parse.

stianjensen:
> Since the behaviour of new Date varies from browser to browser, I'd rather give moment the date in a format that it can definitely parse.

I fully agree with this, that's why I wanted to use Django's default ISO 8601 format. But if this doesn't work for some reason, I'm open to any implementation that ensures moment deterministically parses the date :+1:

dwitter/static/js/scrolling.js
lionleaf: I totally agree, it's getting cumbersome and error prone to deal with. Some bigger refactoring, especially of most of the js code is needed 👍 I'd like to do that refactor myself; but it would most likely have to wait a week or two at least.

PR #190, opened 2017-03-12
Move error div outside the dweet iframe

lionleaf:
Currently only works for the "new dweet" editing box.

screenshot 2017-03-13 00 04 38

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -55,6 +55,11 @@ input[type=submit].delete-button {
   font-family: 'Roboto', sans-serif;
 }
 
+.error-display {
+  color: red;
+  font-size: small;
+  font-family: 'Roboto', sans-serif;
+}
 
 a {
   color: #666;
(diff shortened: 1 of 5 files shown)


tbroadley:
Looks good to me :) Do you want to handle moving errors out of the iframe when remixing in this PR, or in a separate one?

dwitter/static/main.css
```diff
@@ -53,6 +53,11 @@ input[type=submit].delete-button {
   font-family: 'Roboto', sans-serif;
 }
 
+.error-display {
+  color: red;
+  font-size: small;
+  font-family: sans-serif;
```
stianjensen: Please use `font-family: 'Roboto', sans-serif;` like above, for consistency :)

lionleaf: Fixed.

lionleaf:
Now applies to all dweets!
I'll merge and deploy on review :)

lionleaf:
ping

Issue #187, opened 2017-03-12
Javascript linting


tbroadley:
It would be nice to lint our JS, along with our Python. I'd suggest adding the JS linting to Travis as well.

lionleaf:
Fixed with #213

PR #195, opened 2017-03-14
Add Javascript linting (part 1)

tbroadley:
This PR partially completes https://github.com/lionleaf/dwitter/issues/187. It uses ESLint to lint Dwitter's JS files.

It may be easier to review this PR by looking at https://github.com/lionleaf/dwitter/pull/195/files?w=1, a diff with whitespace differences suppressed.

I have another PR ready that resolves the following issues:

- There are still over 100 unresolved linting errors. (This PR only fixes the errors that ESLint can fix automatically.)
- JS linting is not added to the `make lint` command, and thus will not run in Travis.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,0 +1,27 @@
+extends:
+  - defaults/configurations/airbnb/es5
+rules:
+  no-empty-label: off
+  space-after-keywords: off
+  space-return-throw-case: off
+  no-labels: error
+  keyword-spacing: error
+  space-before-function-paren:
+    - error
+    - never
+  comma-dangle:
+    - error
+    - always-multiline
+  no-multiple-empty-lines:
+    - error
+    - max: 1
+  indent:
+    - error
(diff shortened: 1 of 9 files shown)


.eslintrc.yml
```diff
@@ -0,0 +1,27 @@
+extends:
+  - defaults/configurations/airbnb/es5
+rules:
+  no-empty-label: off
+  space-after-keywords: off
+  space-return-throw-case: off
+  no-labels: error
+  keyword-spacing: error
```
tbroadley: The five lines above disable some deprecated rules in the `defaults/configurations/airbnb/es5` config, and replace them with equivalent ones.

Once `eslint-config-defaults` 10.0.0 is released, these lines can be removed.

tbroadley: Alternatively, I might try to switch to using `eslint-config-airbnb-base/legacy` directly from https://github.com/airbnb/javascript.

PR #202, opened 2017-03-16
correctly initialize stats on Firefox

tbroadley:
I found this bug on Firefox 52.0. It seems that Firefox throws an error when trying to render the FPS widget on the hidden new dweet iframe. This change ensures that the FPS widget is rendered correctly.

### Before:

without-stats-fix

### After:

with-stats-fix

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -111,11 +111,15 @@
   <canvas id=c></canvas>
   <div class=error-display></div>
   <script>
-    var stats = new Stats();
-    stats.showPanel(0);
-    stats.dom.setAttribute('id', 'stats');
-    document.body.appendChild(stats.dom);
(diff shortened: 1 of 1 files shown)


dwitter/templates/dweet/dweet.html
```diff
@@ -111,11 +111,15 @@
...
-    document.body.appendChild(stats.dom);
+    function createStats() {
+      var stats = new Stats();
+      stats.showPanel(0);
+      stats.dom.setAttribute('id', 'stats');
+      document.body.appendChild(stats.dom);
+      return stats;
+    }
 
+    var stats = createStats();
```
tbroadley: If `new Stats()` in `createStats` fails, `stats` will be `undefined`. It'll then be reinitialized once the element that contains it is visible, in `loop`.

PR #203, opened 2017-03-16
refresh Waypoints on comment load or new dweet form slide-down

### LINKED ISSUE #194: Autoplay doesn't work properly after long comment threads have loaded
After loading all comments on a long comment thread, scrolling down starts playing the dweets earlier, and stops playing them while they're in the middle of the screen.

To reproduce in prod:
1. Go to top
2. Find a dweet with many comments (for instance https://www.dwitter.net/d/1494)
3. Expand the comment thread.
4. Scroll further down.

tbroadley:
Closes #194.

With this change, [`Waypoint.refreshAll`](http://imakewebthings.com/waypoints/api/refresh-all/) is called whenever the layout of the page is updated (i.e. whenever comments are loaded or the new dweet form slides down).

### DIFF
--- a/dwitter/static/js/ajax-handling.js
+++ b/dwitter/static/js/ajax-handling.js
@@ -49,7 +49,7 @@ var loadComments = function() {
 
   var loadCommentsResponse = function(serverResponse_json, textStatus_ignored,
        jqXHR_ignored) {
-    var $comment_section = $load_comments_button.parents('.comments');
+    var comment_section = $load_comments_button.parents('.comments')[0];
 
     if (serverResponse_json.next) {
       alert('Woops, there are more comments, but they are unloadable as of now. Please bug lionleaf to fix');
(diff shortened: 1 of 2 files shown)


dwitter/static/js/ajax-handling.js
```diff
@@ -64,7 +64,10 @@ var loadComments = function() {
       }
       new_comment_list += getCommentHTML(comment);
     }
-    $comment_section[0].innerHTML = new_comment_list + $comment_section[0].innerHTML;
+    $(comment_section)
+      .html(new_comment_list + comment_section.innerHTML)
+      .promise()
+      .done(Waypoint.refreshAll);
```
tbroadley: Since the comment section's HTML is updated asynchronously, we need to wait until the update is finished before calling `Waypoint.refreshAll`. This is accomplished by converting the call to `html` into a promise, then waiting for the promise to resolve.

PR #206, opened 2017-03-18
Fix black text on new dweet when not logged in

Anheurystics:


Anheurystics:
Something went wrong with the network while I was committing, which introduced some wonky commits. Will make a cleaner version of this later.

Issue #210, opened 2017-04-08
Hot, New, Top and Random


joeytwiddle:
I would really love a "Random" or "Shuffle" page so I can browse through older dweets which I cannot easily reach from the three existing feeds.

Call it "Old" if you want to stick with the three-letter theme. ;)

lionleaf:
I like the idea! Discoverability of old dweets is something that need some work. Searching and hash tags might help too, but this was a simple and effective suggestion. 

PS: a hidden trick to view old dweets:  https://www.dwitter.net/new/page/42   just change 42. Works for hot and top too. It's remaining from before infinite scroll, and might stop working in the future.

joeytwiddle:
That trick is good to know, thanks!

lionleaf:
Implemented in #212, thanks :)

Deployed and looking good

PR #212, opened 2017-04-09
Add a feed page which shows old dweets randomly

joeytwiddle:
Rather than just filing issues, I thought I might try and help out. So here is the "random" feed suggested in #210 .

In this PR I have called it "old", but that could change if you prefer "random".

I used Django's `.order_by('?')`. Because this is not a true shuffle, it carries a small risk of showing the same dweet twice in the feed (inversely proportional to the number of tweets in the system).

I think that is an acceptable payoff, because any approach that performs a true shuffle will also add complexity: either a new field in the DB (which must be updated occasionally) or some kind of temporary state. Better to KISS. Seeing a dweet twice would be a rare occurrence and even then the discomfort felt by the user would be quite small.

I kept the code in line with the other feeds simply for consistency. In fact the page we are on won't make any difference to the results.

I did not provide an "old" feed for user pages, but I add some code for it here and there, in case we want to add that feature in future.

***Needs testing***. Sorry I did not get the environment set up here today. Fortunately the changes I made look pretty simple. But for today at least, I will have to leave it to someone else to check things work as expected.

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -76,6 +76,11 @@ def feed(request, page_nr, sort):
                       .order_by('-hotness', '-posted')[first:last])
         next_url = reverse('hot_feed_page', kwargs={'page_nr': page + 1})
         prev_url = reverse('hot_feed_page', kwargs={'page_nr': page - 1})
+    elif (sort == "old"):
+        dweet_list = (Dweet.objects.annotate(num_likes=Count('likes'))
+                      .order_by('?')[first:last])
+        next_url = reverse('old_feed_page', kwargs={'page_nr': page + 1})
+        prev_url = reverse('old_feed_page', kwargs={'page_nr': page - 1})
     else:
         raise Http404("No such sorting method " + sort)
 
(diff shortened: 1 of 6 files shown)


dwitter/feed/views.py
```diff
@@ -76,6 +76,11 @@ def feed(request, page_nr, sort):
                       .order_by('-hotness', '-posted')[first:last])
         next_url = reverse('hot_feed_page', kwargs={'page_nr': page + 1})
         prev_url = reverse('hot_feed_page', kwargs={'page_nr': page - 1})
+    elif (sort == "old"):
+        dweet_list = (Dweet.objects.annotate(num_likes=Count('likes'))
```
lionleaf: The .annotate() is no longer needed here.

dwitter/user/views.py
```diff
@@ -57,6 +57,9 @@ def user_feed(request, url_username, page_nr, sort, dweets=None, url=None):
     elif (sort == "hot"):
         dweet_list = (dweet_list.annotate(num_likes=Count('likes'))
                       .order_by('-num_likes')[first:last])
+    elif (sort == "old"):
+        dweet_list = (dweet_list.annotate(num_likes=Count('likes'))
```
lionleaf: Annotation can be removed

joeytwiddle:
Enjoying the feature already. Thanks!

PR #214, opened 2017-04-13
Fix last eslint warning and treat warnings as errors

lionleaf:
Is it too much to fail on eslint warnings? Open to a discussion on that point. I'll rebase off the last commit in that case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,11 +27,11 @@ lint-python:
 
 .PHONY: lint-js
 lint-js:
-	npm run lint
+	npm run lint -- --max-warnings 0
 
 .PHONY: lint-js-fix
 lint-js-fix:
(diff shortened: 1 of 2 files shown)


joeytwiddle:
For what it's worth, in my work project I was inspired by standardjs to turn a lot of style conventions into errors with the philosophy:

> If it is a problem that can always be easily fixed (which style changes are), then it should be an error, so that it gets fixed.

This allowed us to retain warnings for things that eslint could not be 100% sure were wrong, but were deserving of developer attention. (Some things like that do exist in our code base.)

----

Having said that, if forbidding warnings is working well on this project, then the code must be very beautiful, and it sounds like a good thing to stick with!

PR #215, opened 2017-04-15
Add CSS for random link when selected (previously overlooked)

joeytwiddle:
Fixes a minor oversight in my previous PR.

Currently when on the random page, the "random" link is not made bold.


lionleaf:
Deployed :)

PR #216, opened 2017-07-27
Update dweet.html

Rybar:
So lo-fi dweets look crispy pixel beautiful!

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -4,6 +4,11 @@
     * {padding:0;margin:0;border:0;outline:0;overflow:hidden}
     canvas {
       width: 100%;
+      image-rendering: optimizeSpeed;
+      image-rendering: -moz-crisp-edges;
+      image-rendering: -webkit-optimize-contrast;
+      image-rendering: -o-crisp-edges;
+      image-rendering: pixelated;
       background: white;
       }
     </style>


joeytwiddle:
***My suggestion: invite Dwitter users to vote this PR up or down with thumb emoticons.***

In most cases I think this would be an improvement, but occasionally the smooth-scaling is used for effect.

Does it make these look better or worse? [#188 by ocasta181](https://www.dwitter.net/d/188), [#1141 elegant](https://www.dwitter.net/d/1141), [#2934 carpet](https://www.dwitter.net/d/2938), [#1283 noisy grid](https://www.dwitter.net/d/1283)

Of course the lower the resolution, the greater effect this change has: [#2928](https://www.dwitter.net/d/2928), [#2944](https://www.dwitter.net/d/2944), [#2932 police](https://www.dwitter.net/d/2932)

It's pretty easy to test this out yourself. Right-click inspect the canvas and paste in the CSS:

It's worth noting that these rules does not affect internal canvas operations, e.g. `transform('scale()')` or [`drawImage`](https://codegolf.tk/a/68), only the way the final raster is displayed.

```
      image-rendering: optimizeSpeed;
      image-rendering: -moz-crisp-edges;
      image-rendering: -webkit-optimize-contrast;
      image-rendering: -o-crisp-edges;
      image-rendering: pixelated;
```

(You need to do it for each canvas individually, because they each live in their own iframe. Ctrl-Z and Ctrl-Y to toggle.)

lionleaf:
Thanks for the PR, @Rybar !

It would be pretty awesome to be able to make more accurate pixel art effects. However, as pointed out by @joeytwiddle, this changes the visuals of a lot of previous dweets, which obviously have been tweaked to look good with the current system. Low res has been used in a lot of dweets for the effect it provides now.

Without those consideration I would prefer this PR to be merged, but I'm not sure as it stands. I'll have to give it a think, and I invite others to voice their opinion! Especially people who have low-res dweets. @sigvef @tbroadley @iver56 @EpicOrange 

I'll be on gitter :)

p01:
Please put this behind a checkbox. 
That's enough to store the state in localStorage, or not store at all even.

But changing the image-rendering of the canvas will break many dweets. The smoothing rendering is often used to get grayscale pixels by drawing a black rectangle with a fractional width and height. It's also often used to get smooth color transitions between colors or when drawing/scaling images.

iver56:
Arguably some dweets would look better when blocky. For example this one:
https://www.dwitter.net/d/448

Other dweets, such as https://www.dwitter.net/d/48 (this one is actually only 64 pixels wide) will not look like intended without the smoothing.

Example of greyscale pixels by drawing sub-pixel rectangles: https://www.dwitter.net/d/89

So yeah, the change is a controversial. Perhaps make it configurable (by the author) per dweet? I agree with p01.

sigvef:
Is it worth potentially breaking three thousand existing dweets for this?

aleksanb:
Backwards compatibility doesn't have to break for such a system.
Each tweet will have to be tagged in the database with an "epoch" or "version" which corresponds to the bootstrap code it was created with.
We then create a new database migration giving all existing tweets the "2016-whenever-dwitter-started" tag , while new tweets get "2017-29-07" or something.
The template will then switch it's bootstrap code depending on  which epoch the dweet is from.

FireyFly:
Forcing crisp rendering for new dweets whilst defaulting to interpolated for old dweets isn't great either, since that means there would exist old dweets that can no longer be reproduced, which doesn't really make it a fair platform.

I think perhaps the easiest and most fair solution would be to have a checkbox for the *viewer* of a dweet to choose to render the dweet crispy (or a button to toggle crispness), which defaults to interpolated just like all dweets are currently.

I think it would be nice to have the option to view lowres dweets with crisp scaling personally (but of course backwards compat is an important concern too).

lionleaf:
I will close this for now. While I like the crisp pixel rendering, it won't go in as a default for now. If there's a "dweet-lang 2.0" at some point, I'll get back to it ;)

PR #218, opened 2017-09-23
Fix CSS wrapping issue on mobile of code block

lionleaf:
When remix text + username got too long for screen it would overflow and
squish the code block. This fixes that issue by playing with the dark
magic that is CSS flow rules. @sigvef / @stianjensen / CSS ninjas out there: I played around with the CSS until it worked, is there anything fundamentally flawed with this technique?

Before:
screenshot 2017-09-24 00 37 03

After:
screenshot 2017-09-24 00 37 37

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -238,6 +238,8 @@ input.comment-submit:active {
 
 .author-remix-wrapper{
   padding: 10px 0;
+  display: inline-block;
+  width: 100%;
 }
 
 .dweet-author {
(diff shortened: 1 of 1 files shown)


dwitter/static/main.css
```diff
@@ -238,6 +238,8 @@ input.comment-submit:active {
 
 .author-remix-wrapper{
   padding: 10px 0;
+  display: inline-block;
```
stianjensen: This works, and I don't think it will have any unintended quirks, although a more common way of resolving this issue is applying `overflow: hidden`, or using a `clearfix` hack (https://www.w3schools.com/howto/howto_css_clearfix.asp)

PR #222, opened 2017-10-30
Make Dwitter header fixed, move 'new dweet' button

stianjensen:
I tried looking at tweaking the Dwitter header to make the buttons in the header always accessible even when you've scrolled miles down the page. To make this work I decided to reduce the height of the header somewhat to make it not take up not much permanent vertical space.

I also moved the 'New dweet' button into the header, so you don't have to scroll all the way up to find it.

If I didn't mess up, this should also work pretty well on mobile.

Screenshots:
screen shot 2017-10-30 at 19 38 19
screen shot 2017-10-30 at 19 38 16
screen shot 2017-10-30 at 19 38 02

### DIFF
--- a/dwitter/static/js/feed.js
+++ b/dwitter/static/js/feed.js
@@ -68,7 +68,6 @@ $(document).ready(function() {
     registerWaypoint(iframe);
     /* eslint-enable no-undef */
 
-    $(this).hide();
     $(dweet).find('textarea').focus();
   });
 });
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -115,7 +115,7 @@ ul.messages {
 }
 
 #content {
-  padding: 0 0 1em;
+  padding: 50px 0 1em;
   margin: 0 auto;
 }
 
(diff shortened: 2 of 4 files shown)


joeytwiddle:
My mobile is 480x854. At such resolutions, the permanent header is somewhat problematic for unregistered users, and very problematic in landscape mode!

### Portrait (logged out vs logged in)

 . . . . . 

### Landscape (logged out vs logged in)

 . . . . . 

## Possible solutions

1. On mobile, revert to the old behaviour: do not make the header permanent.

2. On mobile/everywhere, hide the header when scrolling down, show when scrolling up. (**My favourite**)

3. On mobile, hide the header behind a burger menu: ☰ (but that will still take some space!)

stianjensen:
I think the 'login or register' link could be changed to a simple 'login' link. That will make the header less busy on both mobile and desktop.

joeytwiddle:
That will help, but that won't change the last screenshot above. I would really like to scroll through dweets in landscape mode. Even if we compress it down a lot, the header will prevent me from seeing the whole dweet canvas on screen at once!

PR #223, opened 2017-10-30
Upgrade Django from 1.8.x to 1.11

stianjensen:
Django 1.11 is the next LTS release of Django and the last release to
support Python 2. Luckily, Django 1.11 will be supported with security
updates until sometime in 2020 so there's no immediate rush to rewrite
everything to Python 3 just yet.

Still, Django 1.8 will soon stop getting security updates, so upgrading
to Django 1.11 now means it will still be easy to take in new Django
Security fixes.

Migrating was fairly straight-forward, and involved reading the release
notes of Django 1.9, 1.10 and 1.11, as well as upgrading any third-party
libs that had released compatibility updates for Django 1.11:

- https://docs.djangoproject.com/en/1.11/releases/1.9/
- https://docs.djangoproject.com/en/1.11/releases/1.10/
- https://docs.djangoproject.com/en/1.11/releases/1.11/

The biggest problem was related to managing subdomains, so I had to
switch to a fork of the subdomain library that had added support for
Django's new middleware style. Also, I would double check the
ALLOWED_HOSTS setting in Production before deploying (although I suspect
it must already be set to something valid). Anyway, I had some issues
where it seems you also need to specify allowed_hosts locally now.


stianjensen:
fwiw, it should be trivial to upgrade to the latest 1.8.x security release from march this year (1.8 had end of support april this year)

https://docs.djangoproject.com/en/2.0/releases/1.8.19/

PR #226, opened 2017-12-27
Caching

sigvef:


dwitter/templates/feed/feed.html
```diff
@@ -59,9 +60,17 @@
...
-  {% for dweet in dweet_list %}
-    {% include 'snippets/dweet_card.html' %}
-  {% endfor %}
+  {% if request.user.is_authenticated %}
+      {% for dweet in dweet_list %}
+        {% include 'snippets/dweet_card.html' %}
+      {% endfor %}
+  {% else %}
+      {% for dweet in dweet_list %}
+        {% cache 900 dweetcard dweet.pk %}
```
lionleaf: Is 900 seconds (15 min) cache on each individual dweet a bit too much? Granted, it's just for people who haven't logged in, and it will still display new dweets on time. It should only affect Awesome-count and comments.
Works well when we're under heavy load, but for normal operations we might want to turn it down?

But to how much?

dwitter/dweet/views.py
```diff
@@ -13,6 +14,7 @@ def handler404(request):
 
 
 @xframe_options_exempt
+@cache_page(3600)
```
lionleaf: As there's currently no way to change dweets themselves, this cache shouldn't be a problem for now. Just have to make sure to remember this in the future.

lionleaf:
60s sounds pretty reasonable, especially since this is only for people who haven't logged in.

sigvef:
rebased

PR #234, opened 2017-12-27
Add flake8 to testing in both Python 2 and 3

cclauss:


sigvef:
I think we already run flake8 through `make lint-python` (: Nice catch to add python 3 to the travis build, though!

Issue #173, opened 2017-03-06
Feature request: turn dwitter URLs in comments into links.


veu:
It would be great if dwitter URLs in comments would be converted to links.

Example input: `Look at https://www.dwitter.net/d/1234`
Example output: `Look at <a href="/d/1234">#1234</a>`

tbroadley:
This is definitely something we should consider doing when implementing https://github.com/lionleaf/dwitter/issues/150.

For now, we do turn URLs in comments into links: https://github.com/lionleaf/dwitter/blob/2d9ea80e632ad4cb06c5cf3166603f668f55f889/dwitter/templates/snippets/dweet_card.html#L92

This isn't perfect, though. A newly-added comment won't have its URLs turned into links until after refreshing the page.

atesgoral:
#244 with PR #245

atesgoral:
@tbroadley Can't repro the "until refresh" issue. This might have been addressed already. See: https://github.com/lionleaf/dwitter/blob/master/dwitter/serializers.py#L23

tbroadley:
@atesgoral You're right! Looks like that issue was tracked in https://github.com/lionleaf/dwitter/issues/184, and was fixed in #204. Thanks for pointing that out.

lionleaf:
fixed

atesgoral:
@lionleaf Actually, can we reopen this? I think @veu wasn't talking about what we've just implemented. It's about also shortening full URLs to the magic pattern, which is something I was hoping to land in a later PR. This is a nice feature to have.

lionleaf:
Woops, I was a bit too quick. Reopened

tbroadley:
Should this issue be closed (because of cc5be1b)?

veu:
Sure 👍

Issue #150, opened 2017-02-28
Mention users and dweets in comments


lionleaf:
Being able to mention other users and dweets in comments in a way that automatically links. I'm planning to use the following syntax:

u/username for a user and d/1337 for dweets.  (or /d/1337 and /u/username?). 

I'm avoiding @ mentions because @ is a valid symbol in usernames (defaults in django). 
And #, while nice for dweet-mentions, could potentially be used for topic-tags at some point in the future. ( #oslo-dweet-compo-17 anyone?)

sigvef:
Remember autocomplete! Also, d/1337 is better than /d/1337.

atesgoral:
#244 with PR #245

lionleaf:
fixed by #245

PR #245, opened 2018-01-02
insert_magic_links filter for dweet and user links

### LINKED ISSUE #244: Magic links for dweet and author mentions
Just like what GitHub does for commit hashes, automatically convert certain patterns to dweet or author links:

* /u/lionleaf -- user link
* /d/123 -- dweet link

atesgoral:
Resolves #244 by introducing a new template filter. This could have been also done as a template tag, but I followed to_gravatar_url (which is a filter inside the templatetags folder) as an example.

Potential improvements:
* [x] Make the regexp more specific to ignore cases where the patterns appear in the middle of a URL etc.

### DIFF
--- a/dwitter/serializers.py
+++ b/dwitter/serializers.py
@@ -1,5 +1,6 @@
 from rest_framework import serializers
 from dwitter.models import Dweet, Comment
+from dwitter.templatetags.insert_magic_links import insert_magic_links
 from django.contrib.auth.models import User
 from django.template.defaultfilters import urlizetrunc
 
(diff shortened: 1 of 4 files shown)


atesgoral:
I think it the pattern should be either d/n or /d/n, but not both. What do you think? And which pattern do you prefer?

lionleaf:
Nice! I like the use of template tags. Works well locally. Happy to merge and deploy once the linter is happy ^^ 

PS: We've got travis set up to be pretty picky about formatting, so a pro-tip is to run `make lint` locally first :)

lionleaf:
To answer @joeytwiddle :   I want to save #tags for a future hashtag implementation, not for dweet links.

full url compression sounds like a nifty addition though! :D

lionleaf:
fixes #173 and #244 

(that should link the issues properly)

stianjensen:
> fixes #173 and #244 and #150

> (that should link the issues properly)

Heads up, this will only close the first of the three issues 😅

lionleaf:
fixes #244 
fixes #150

PR #246, opened 2018-01-03
Make leading forward slash optional

atesgoral:
Both /u/a and u/a are supported. Display text is normalized to u/a to encourage that flavour.


sigvef:
thanks :)

lionleaf:
published!

PR #251, opened 2018-01-06
Remix info area fixes

### LINKED ISSUE #250: "remix of #n by username" text misleading about the way to link to dweets
Probably due to the "#n" text in remixed dweets, people have picked up the convention of referring to other dweets with the #n syntax. Some ideas:

* Change the remix text to say "remix of d/n by username" or even "remix of d/n by u/username" (and maybe make the d/ and u/ parts slightly faded out) as a hint to users on how to refer to other dweets/users -- but that might look ugly
* Also accept the #n format as a way to link to dweets -- this should be a problem for future #tags since tags will very likely not be numeric.

sigvef:
> **Fix [deleted] message in dweet remix info area** 
Before, it would look like the user of the remixed dweet was deleted,
but really, it's the dweet itself that was deleted. Now, it looks more
like the latter.

> **Use proper d/ and u/ style in remix info area**
This can work as a way of letting users know what the syntax is for
linking to dweets and users in the comments.
>
>This fixes #250.

### DIFF
--- a/dwitter/templates/snippets/remix_info.html
+++ b/dwitter/templates/snippets/remix_info.html
@@ -1,11 +1,11 @@
 {% load to_gravatar_url %}
 <div class="remix-info">
   remix of
-  <a href="{% url 'dweet_show' dweet_id=dweet.reply_to.id %}">#{{dweet.reply_to.id}}</a>
-  by
   {% if dweet.reply_to.deleted %}
     [deleted]
   {% else %}
+    <a href="{% url 'dweet_show' dweet_id=dweet.reply_to.id %}">#{{dweet.reply_to.id}}</a>
+    by
     <a href="{% url 'user_feed' url_username=dweet.reply_to.author.username %}">
       {{ dweet.reply_to.author.username }}
     </a>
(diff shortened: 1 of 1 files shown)


lionleaf:
screen shot 2018-01-06 at 12 23 49 pm
From prod :D

dwitter/templates/snippets/remix_info.html
```diff
@@ -1,13 +1,13 @@
 {% load to_gravatar_url %}
 <div class="remix-info">
   remix of
-  <a href="{% url 'dweet_show' dweet_id=dweet.reply_to.id %}">#{{dweet.reply_to.id}}</a>
-  by
   {% if dweet.reply_to.deleted %}
     [deleted]
   {% else %}
+    <a href="{% url 'dweet_show' dweet_id=dweet.reply_to.id %}">d/{{dweet.reply_to.id}}</a>
```
atesgoral: Could you put the d/ and u/ in a `<span class="faded"></span>` (`.faded { opacity: .75 }`)?

PR #252, opened 2018-01-06
Change dweet card u function syntax to be consistent with new dweet

atesgoral:
Missed a spot while doing this change earlier.


sigvef:
Nice catch!

Issue #254, opened 2018-01-07
API to access dweets in JSON


lionleaf:
Requested feature to be able to get dweets as JSON. 

I'll implement something basic, let me know if people have specific requirements / suggestions for the API

PR #255, opened 2018-01-07
Add API to get JSON dweets and remixes

lionleaf:
/api/d/123  --> Dweet 123 in JSON

Example JSON response:

`{"id":4054,"code":"c.width=1920; /* clear the canvas */\r\n  for(i=0;i<9;i++)\r\n  x.fillRect(400+i*100+S(t)*300, 400, 50, 200) /* draw 50x200 rects test */","posted":"2017-12-30T23:19:57.490518Z","author":1,"author_name":"lionleaf","awesome_count":1,"remix_of":4053}`

/api/newest --> Newest dweet

JSON looks the same as for /api/d/XXX but will always show only the newest one (good for polling for new dweets)

/api/d/123/remixes --> all remixes of d/123

Thoughts? Is this the right way to do it? Or maybe for a more generic /api/d/  <--- list of all dweets and then add ?parameter= filters so you can do your own queries? Slightly worried about performance, but guess we can rate limit

### DIFF
--- a/dwitter/serializers.py
+++ b/dwitter/serializers.py
@@ -6,19 +6,31 @@ from django.template.defaultfilters import urlizetrunc
 
 
 class UserSerializer(serializers.ModelSerializer):
+    link = serializers.SerializerMethodField()
+    date_joined = serializers.DateTimeField(format="%Y-%m-%d")
+
     class Meta:
-        id = serializers.ReadOnlyField(source='pk')
         model = User
-        fields = ('id', 'username')
+        fields = ('username', 'date_joined', 'link')
+
+    def get_link(self, obj):
+        return 'https://www.dwitter.net/u/%s' % obj.username
 
 
 class CommentSerializer(serializers.ModelSerializer):
-    author = serializers.ReadOnlyField(source='author.username')
     urlized_text = serializers.SerializerMethodField()
+    author = serializers.SerializerMethodField()
+    reply_to = serializers.PrimaryKeyRelatedField(
+        queryset=Dweet.with_deleted.all()
+    )
+    id = serializers.ReadOnlyField(source='pk')
 
     class Meta:
         model = Comment
-        fields = ('pk', 'urlized_text', 'text', 'posted', 'reply_to', 'author')
+        fields = ('id', 'urlized_text', 'text', 'posted', 'author', 'reply_to')
+
+    def get_author(self, obj):
+        return obj.author.username
 
     def get_urlized_text(self, obj):
         return insert_magic_links(urlizetrunc(obj.text, 45))
@@ -30,16 +42,17 @@ class DweetSerializer(serializers.ModelSerializer):
         source='reply_to',
         queryset=Dweet.with_deleted.all()
     )
+    link = serializers.SerializerMethodField()
     awesome_count = serializers.SerializerMethodField()
-    author_name = serializers.SerializerMethodField()
+    author = UserSerializer()
 
     class Meta:
         model = Dweet
-        fields = ('id', 'code', 'posted', 'author', 'author_name', 'awesome_count',
+        fields = ('id', 'code', 'posted', 'author', 'link', 'awesome_count',
                   'remix_of')
-
-    def get_author_name(self, obj):
-        return obj.author.username
 
     def get_awesome_count(self, obj):
         return obj.likes.all().count()
+
+    def get_link(self, obj):
+        return 'https://www.dwitter.net/d/%i' % obj.id
--- a/dwitter/serializers.py
+++ b/dwitter/serializers.py
@@ -7,34 +7,39 @@ from django.template.defaultfilters import urlizetrunc
 
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
+        id = serializers.ReadOnlyField(source='pk')
         model = User
-        fields = ('pk', 'username')
+        fields = ('id', 'username')
 
 
 class CommentSerializer(serializers.ModelSerializer):
     author = serializers.ReadOnlyField(source='author.username')
-    posted = serializers.ReadOnlyField()
     urlized_text = serializers.SerializerMethodField()
 
     class Meta:
         model = Comment
-        fields = ('urlized_text', 'text', 'posted', 'reply_to', 'author')
+        fields = ('pk', 'urlized_text', 'text', 'posted', 'reply_to', 'author')
 
     def get_urlized_text(self, obj):
         return insert_magic_links(urlizetrunc(obj.text, 45))
 
 
 class DweetSerializer(serializers.ModelSerializer):
-    latest_comments = serializers.SerializerMethodField()
-    reply_to = serializers.PrimaryKeyRelatedField(
+    id = serializers.ReadOnlyField(source='pk')
+    remix_of = serializers.PrimaryKeyRelatedField(
+        source='reply_to',
         queryset=Dweet.with_deleted.all()
     )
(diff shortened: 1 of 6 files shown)


sigvef:
Perhaps we should nest the author as a full object, e.g.

```
{"id":4054,"code":"..","author: {"id":1, "username":"lionleaf", ..},"awesome_count":1,"remix_of":4053}
```

sigvef:
serving on api.dwitter.net is probably a good idea

sigvef:
Since you asked about URLs and such: If I were to design the urls it might have been something like this:

```
GET /dweets/ - paginated list of dweets, ordered by publish date by default
GET /dweets/?author=123 - paginated list of dweets filtered by author
GET /dweets/123 - single dweet with id 123
GET /dweets/123/remixes/ - list of dweets that have remixed the dweet with id 123
```

using http://www.django-rest-framework.org/api-guide/filtering/ for the url param filters

sigvef:
(And later: `POST /dweet/` creates a new dweet)

atesgoral:
I like @sigvef's `GET /dweets/*` path suggestions.

The canonical dweet and user URLs could also be included in the response so that API consumers don't have to construct the URLs (which might be subject to change) themselves.

P.S. GraphQL is all the rage nowadays.

atesgoral:
Also, I'd suggest manually-managed API keys, just in case...

atesgoral:
BTW :)

https://dweetplayer.net/api/dweets/1994

sigvef:
> The canonical dweet and user URLs could also be included in the response so that API consumers don't have to construct the URLs (which might be subject to change) themselves.

Good idea

sigvef:
> Also, I'd suggest manually-managed API keys, just in case...

Might be able to get away with basic ratelimiting on IP (which happens to be built-in to DRF :) )

lionleaf:
Thanks guys 👍  Lots of good input, I'll do another revision. Keep it coming!

lionleaf:
What are the arguments for api.dwitter.net vs dwitter.net/api ?

atesgoral:
I don't see an immediate use for a separate subdomain. `/api` could do just fine.

On a tangential topic, there's a way to run stuff inside an iframe without needed an actual subdomain. So, subdomains could be entirely done away with. I applied that technique at https://canvas.run (unfinished, unpublicized pet project of mine): You can load an iframe through a base64 data URL which contains a simple bootstrap that receives the main iframe runtime via postMessage. Browsers (the modern ones I tested with) treat iframes with data URLs as being of a different origin and therefore the same restrictions as having an actual different subdomain apply.

sigvef:
> You can load an iframe through a base64 data URL which contains a simple bootstrap that receives the main iframe runtime via postMessage. Browsers (the modern ones I tested with) treat iframes with data URLs as being of a different origin and therefore the same restrictions as having an actual different subdomain apply.

that's pretty neat!

lionleaf:
[DON'T MERGE YET]

Pushed my current WIP to keep the conversation going. Have some minor tweaks I want to get in before a merge. Also have to verify that posting comments still work.

Current API:


```
GET /dweets/          - list of the last 10 dweets
       ?limit=100        - number of results to return, default 10, max 100
       ?offset=200     - offset page by 200 dweets
       ?reply_to=123  - all remixes of 123
       ?author=1         - dweets by author (think it's authorID atm, will change to username)
GET /dweets/123/remixes/ - list of dweets that have remixed the dweet with id 123.  (duplicate API atm)
GET /dweets/newest/ - newest dweet
```

```
GET /dweets/123 - single dweet with id 123
<img width="1054" alt="screen shot 2018-01-09 at 10 19 55 pm" src="https://user-images.githubusercontent.com/610925/34758918-f7ca98ce-f58d-11e7-8f72-c8476585d4fe.png">
```

```
GET /users/lionleaf
<img width="457" alt="screen shot 2018-01-09 at 10 25 27 pm" src="https://user-images.githubusercontent.com/610925/34758978-44d35e94-f58e-11e7-8789-c94e18a735b5.png">
```

Not possible to list _ALL_ users at the moment, do we want people to see the list of all users? What about the internal user ID? I've currently chosen to not show it.

Will switch off API explorer in prod; JSON only (API explorer does some extra database-queries that seem impossible to remove, consuming the CPU for almost a minute if you GET /comments/)

atesgoral:
* reply_to - could this be called remix_of?
* urlized_text - could this be called url? - also it's going to be a pure URL, without the anchor tag, right?
* /dweets/newest/ - could this be handled as /dweets?sort_by=posted&sort_dir=desc
* limit, offset - another common convention is to go with page_size, page_num

lionleaf:
> reply_to - could this be called remix_of?

As the url parameter? Yes. I think it should remain reply_to for the comment API tho :P

> urlized_text - could this be called url? - also it's going to be a pure URL, without the anchor tag, right?

Do you mean renaming `link` to `url`? urlized_text is the whole comment but with automatic <a href> tags. Again, I think I might've confused things by including a screenshot from the comments API.  The user and dweet API has a full/permalink called link at the moment. Happy to rename to `url`, but I find `link` to be equally good

> /dweets/newest/ - could this be handled as /dweets?sort_by=posted&sort_dir=desc
Yes, you're right. Both /newest and /remixes is now redundant. I'll remove them, and rather show how to get it in the README.

> limit, offset - another common convention is to go with page_size, page_num
I'm using limit, offset for the comment API, in particular because of the offset when fetching comments, where 3 comments have been prefetched.
In general page_size + page_num might be more common, but unsure if I want to mix and match between the APIs.



PS: I know some people are waiting for this, so apologies for the speed. Been busy with life. Will most likely get back to it in the middle of next week I'm afraid (going away for the weekend! :D )

lionleaf:
This fixes #254 

(just linking the issue)

dwitter/serializers.py
```diff
@@ -6,35 +6,53 @@
...
+    link = serializers.SerializerMethodField()
+    date_joined = serializers.DateTimeField(format="%Y-%m-%d")
+
     class Meta:
         model = User
-        fields = ('pk', 'username')
+        fields = ('username', 'date_joined', 'link')
+
+    def get_link(self, obj):
+        return 'https://www.dwitter.net/u/%s' % obj.username
```
atesgoral: I was using the urlresolver in the insert_magic_links filter, just to be safe, but hard-coding the URL should be fine for all intents and purposes :)

dwitter/serializers.py
```diff
@@ -6,35 +6,53 @@
...
+                  'remix_of')
+
+    def get_awesome_count(self, obj):
+        return obj.likes.all().count()
 
-    def get_latest_comments(self, obj):
-        cmnts = obj.comments.all().order_by('-posted')
-        return CommentSerializer(cmnts[:3], many=True).data
+    def get_link(self, obj):
+        return 'https://www.dwitter.net/d/%i' % obj.id
```
atesgoral: Consider using urlresolver?

lionleaf:
Updated API:

### Dweets
```
GET dwitter.net/api/dweets/  - list of the last 10 dweets
       ?limit=100            - number of results to return, default 10, max 100 (subject to change)
       &offset=200           - offset page by 200 dweets
       &reply_to=123         - all remixes of 123
       &author=lionleaf      - dweets by author
```

Latest dweet: `https://www.dwitter.net/api/dweets/?limit=1`  (sorted by posted date by default)


### Users
```
GET dwitter.net/api/users/lionleaf  - Show details about user 'lionleaf'.
```

atesgoral:
💯

PR #256, opened 2018-01-08
Roll up anchor content to remove extra space

atesgoral:
Currently, the anchor hangs beyond the username:

image

With this fix:

image


lionleaf:
Nice catch indeed!

Deployed 👍

PR #257, opened 2018-01-13
add link to liked on user page

whackashoe:
add link to awesome for user pages

this should probably have a route which ignores posts from owning user

### DIFF
--- a/dwitter/templates/feed/feed.html
+++ b/dwitter/templates/feed/feed.html
@@ -36,6 +36,7 @@
   {%elif feed_type == "user" %}
   <li><a class="new" href="{% url 'user_sort_feed' sort='new' url_username=feed_user %}">new</a></li>
   <li><a class="top" href="{% url 'user_sort_feed' sort='top' url_username=feed_user %}">top</a></li>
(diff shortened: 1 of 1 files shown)


dwitter/templates/feed/feed.html
```diff
@@ -36,6 +36,7 @@
   {%elif feed_type == "user" %}
   <li><a class="new" href="{% url 'user_sort_feed' sort='new' url_username=feed_user %}">new</a></li>
   <li><a class="top" href="{% url 'user_sort_feed' sort='top' url_username=feed_user %}">top</a></li>
+  <li><a class="awesome" href="{% url 'user_liked' sort='new' url_username=feed_user %}">liked</a></li>
```
lionleaf: I'd prefer to not use "like"-terminology as we don't have a "like" button per se. However, I'm not sure "awesomed" is any better. @sigvef thoughts?

whackashoe:
So if you have a page of liked posts it will include all posts the user put up... but thats probably not what we want, we'd want to filter those out to just see what they've clicked the button on. Since we just came from their profile page we'd already know what they've dweeted.

lionleaf:
Riiight, because you automatically awesome your own posts

PR #258, opened 2018-01-13
add about page stub

whackashoe:


whackashoe:
For everyone reading: please suggest:

*  wording
* layout of about page
* link position in menu

Or whatever else you notice :)

lionleaf:
Thanks!
I'm going to re-open this from a local branch, so multiple people can add commits to it. Now that you're a collaborator, you should set up your repository to point here instead of your fork and you can create local branches yourself :) 

(never done this before on github, but I suspect this is the best way to do it)

PR #260, opened 2018-01-18
Restrict API results to only JSON for performance

lionleaf:
The django API explorer does a ton of extra database requests, which
makes any usage of it on a decent size dataset a DoS attack ^^ At least
with my current settings.


iver56:
Alternatively, this can be moved to `local.py` that resides in the production environment, because we don't need to disable the BrowsableAPIRenderer in all dev environments

PR #268, opened 2018-01-20
Awesome counts

### LINKED ISSUE #265: Awesome-count is broken in multiple places
Broken on both /random and u/lionleaf/awesome (and I think that should be it)

@sigvef

sigvef:
This fixes #265 .

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -78,8 +78,7 @@ def feed(request, page_nr, sort):
         next_url = reverse('hot_feed_page', kwargs={'page_nr': page + 1})
         prev_url = reverse('hot_feed_page', kwargs={'page_nr': page - 1})
     elif (sort == "random"):
-        dweet_list = Dweet.objects.annotate(
-            num_likes=Count('likes')).order_by('?')[first:last]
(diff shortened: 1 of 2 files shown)


dwitter/feed/views.py
```diff
@@ -92,6 +91,12 @@ def feed(request, page_nr, sort):
         .select_related('reply_to__author__username')
         .prefetch_related('comments'))
 
+    # For some reason order_by('?') and .annotate(num_likes=Count('likes'))
+    # don't work together, so we need to do extra work if the sorting is random.
```
lionleaf: This line is a single character too long, so the lint and build fails. FIX IT! :)

iver56: The 80 character limit should not be a hard limit IMO. Fun fact: The 80 character "limit" is based on the legacy of the 80 column punched card format

lionleaf: While I tend to agree I also think it's alright to have a strict build system based on the linter. We could change the lint config though

sigvef:
I reformatted the comment that tripped up the build. Since you've already approved and there's no other changes, I'll go ahead and click merge on this :)

Issue #247, opened 2018-01-04
Code blocks in comments


atesgoral:
Proposal: Use backtick (<code>\`</code>).

* Escape backticks within code as <code>\\`</code>

sigvef:
I like it!

whackashoe:
would hljs (https://highlightjs.org/usage/) be welcome for this?

atesgoral:
Let's hold off on merging this. I'm not 100% sold on the idea of using `` \` ``. Adding support for both ` `` ` and ` ``` ` might be better.

atesgoral:
Syntax highlighting would also be cool if it doesn't turn out to be too expensive computationally.

lionleaf:
oops, I merged straight away ^^ Didn't see your request.

``` might work well.

I would imagine syntax highlighting would be done on the client side with js anyway, so shouldn't be too expensive?

atesgoral:
@lionleaf LOL I accidentally put the comment on the issue instead of the PR. We can enhance later with new PRs.

atesgoral:
Yeah, we can throw in a client-side syntax highlighter.

lionleaf:
Implemented by #270 .  

Code highlighting and other follow-up work can be opened as new issues if you want @atesgoral :D

PR #270, opened 2018-01-20
Backticks for code blocks

atesgoral:
Converts stuff between `` ` `` and `` ` `` to `<code>` blocks, except when a `` ` `` is escaped with a `` \` ``

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -196,6 +196,10 @@ ul.messages li.success{
   word-wrap: break-word;
 }
 
+.comment-message code {
+  background: #eee;
+}
+
 .new-comment {
   margin-top: 10px;
   display: flex;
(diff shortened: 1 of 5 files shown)


atesgoral:
Closes #247

dwitter/static/main.css
```diff
@@ -196,6 +196,10 @@ ul.messages li.success{
   word-wrap: break-word;
 }
 
+.comment-message code {
+  background: #eee;
```
lionleaf: Might want to change font too?

atesgoral: The font is already set by the top-level `code` CSS rule. Same font as the editor. Might want to play around with adding a border + padding around, but then layout might get thrown off.

lionleaf: Oh, right. Looks pretty good already :)

lionleaf:
live!

PR #275, opened 2018-01-27
Extended session length to 100 years

srmcgann:


lionleaf:
Though, that's technically slightly less than 100 years :P Not that it matters.

diceroll123:
I believe that's the leap days that bring it down.

PR #276, opened 2018-01-28
Add support for embedding dweets on external sites

### LINKED ISSUE #220: [Feature Suggestion] Dwitter Embed
Could use Dwitter.net/e/#### for embed on external sites just like youtube embed.
Once requested it fetch's the dweets code and responds with standalone player...

`<canvas id="c"width="1920"height="1080"style="width:100%"><script>c=document.body.childNodes[0];with(Math)S=sin,C=cos,T=tan;function R(r,g,b,a){a=a===undefined?1:a;return "rgba("+(r|0)+","+(g|0)+","+(b|0)+","+a+")"};x=c.getContext("2d");frame=0;function u(){t=++frame/60;
//dweet below
for(i=c.width|=0;i--;x.fillRect(960+S(p=1.047*i+S(t/5)*3e3/i*S(t))*(d=i*i/1e3)-i*C(t),540+C(p)*d-i*S(t),s,s))x.fillStyle=R(i*t%99,i/t,s=i/5)
//
requestAnimationFrame(u)}u()</script>`

lionleaf:
Adds a new url /e/DWEET_ID that is a bare-bones display of the dweet
without comments, this can be easily embedded on external sites
(cross-origin policy allows it too!) in an iframe.

ie: <iframe width=500 height=550 frameBorder="0" src="https://www.dwitter.net/e/123">
</iframe>

(width and height is flexible)

(ignore the DjDT in the picture, it's only for debugging)
screen shot 2018-01-28 at 12 51 51 am

embed link added to all dweets (and in the process I moved the fullscreen button to the very right, even when you can see the delete button)

screen shot 2018-01-27 at 11 45 03 pm
screen shot 2018-01-27 at 11 45 07 pm

fixes #220 

(there are some minor styling issues with the error text and size of the iframe)
screen shot 2018-01-28 at 12 52 18 am

### DIFF
--- a/dwitter/feed/urls.py
+++ b/dwitter/feed/urls.py
@@ -36,5 +36,8 @@ urlpatterns = [
         views.dweet_delete, name='dweet_delete'),
     url(r'^d/(?P<dweet_id>\d+)/like$', views.like, name='like'),
 
+    url(r'^e/(?P<dweet_id>\d+)$',
+        views.dweet_embed, name='dweet_embed'),
+
     url(r'^dweet$', views.dweet, name='dweet'),
 ]
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -1,6 +1,7 @@
 from django.shortcuts import get_object_or_404, render
 from django.views.decorators.http import require_POST
 from django.views.decorators.csrf import csrf_protect
+from django.views.decorators.clickjacking import xframe_options_exempt
 from django.http import HttpResponseRedirect, HttpResponse
 from django.http import Http404, HttpResponseBadRequest
 from django.core.urlresolvers import reverse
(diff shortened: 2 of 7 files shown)


atesgoral:
My vision for an embed feature was a bare-bones canvas embed feature to be used inline when writing articles on Medium etc. The source could optionally dangle at the bottom, but then it would make the embedded content not always look good within the context (unless we also add support for dark/light themes etc.). The source could also be only shown when a "show source" button is clicked on a toolbar (that could appear on hover), thinking that in an article context, you'd have the source already in a separate code block. I'd also try not to include any of the waypoint, jQuery, Roboto font, etc. stuff to make this as lightweight as possible.

lionleaf:
@atesgoral I'd much rather have the general embed feature include the author / code / remix credits etc. Even though that might not be exactly what you want for the post you have in mind.

Is this enough for the feature you want?

`<iframe width=568 height=319.5 frameBorder="0" src="https://dweet.dwitter.net/id/6494?autoplay=1"></iframe>`

Tested locally and seems to work just fine already:

screen shot 2018-01-29 at 5 01 18 pm

Should be good for a post that wants to explain different techniques, but if there's a blog post highlighting their favorite dweets, wouldn't you want it to look more like:

screen shot 2018-01-29 at 5 03 22 pm

PR #277, opened 2018-02-03
Fix discord invite links

lionleaf:


lionleaf:
Thanks to @Apex on discord for the report :)

PR #280, opened 2018-02-04
Show New Dweet button on every page

### LINKED ISSUE #237: Add "New Dweet" Button to all pages.

diceroll123:
Fixes #237 

*well that was easy*


diceroll123:
Knew it was too easy to be true. 🤔 

Will try some things.

diceroll123:
Updated as per requested.

*My* only gripe is the differing widths here:

PR #283, opened 2018-02-05
Fix profile info to show when user has 0 dweets

diceroll123:
base.html is now never directly rendered. This was easier than making base.html include all of the static JS files.

---

If anyone wants to ask, `str(total_awesome or 0)` because `total_awesome` returns `None` if there are 0.


diceroll123:
A (positive, in my opinion) side effect from this is that when you're in any dweet feed (be it a profile's hot/top/etc or the regular hot/top/etc) and it has a length of 0, it will say "No dweets here" instead of reaching the last page, which it will now do otherwise.

PR #285, opened 2018-02-05
Fix magic links

### LINKED ISSUE #273: Characters after u/username prevent magic link insertion
Probably should make the regex a bit more forgiving.

Live example: Punky's comment on https://www.dwitter.net/d/6281

>Based on u/sadropol's Christmas tree.

diceroll123:
Fixes #273

Passes all tests, adds 2 tests. Hopefully this'll suffice for now, to be replaced with an even more buffed one eventually.

### DIFF
--- a/dwitter/templatetags/insert_magic_links.py
+++ b/dwitter/templatetags/insert_magic_links.py
@@ -10,25 +10,24 @@ def to_link(m):
     username = m.group('username')
 
     if username is None:
-        path = '/d/' + dweet_id  # hardcode for speed!
-        # path = reverse('dweet_show', kwargs={'dweet_id': dweet_id})
+        url = 'd/' + dweet_id
     else:
-        path = '/u/' + username  # hardcode for speed!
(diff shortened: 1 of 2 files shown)


diceroll123:
Ah, this won't work for things like `u/steve's dweet`

Hmm... 🤔

diceroll123:
The apostrophe thing is an easy fix, but then a bunch of tests will fail, because it would link as much of the value as it matches, per word, instead of passing if it's not precise.

This is not my design decision to make. 😔  For now, this will do. 👍

diceroll123:
The example would be to do with suffixes, not exclusive to apostrophes. It would be like this:

| Input | Output |
|-|-|
| `u/a/suffix` | `<a href="u/a">u/a</a>/suffix` |
| `u/steve's dweet` | `<a href="u/steve">u/steve</a>'s dweet` |

Though I approve of the latter, the former is my only real gripe of making this work. 🤔 A handful of tests will have to be scrapped.

PR #290, opened 2018-02-14
New Function: (L)

Ubiw:
The L function is just like R but it does HSLA instead. I think this could save alot of chars for even more cool styles!!!

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -131,6 +131,10 @@
       a = a === undefined ? 1 : a;
       return "rgba("+(r|0)+","+(g|0)+","+(b|0)+","+a+")";
       };
+    function L(h,s,l,a) {
+      a = a === undefined ? 1 : a;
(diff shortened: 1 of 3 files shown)


lionleaf:
Thanks for the pull request; always appreciated.

However, you hit straight on a contentious issue. While I think the hsla function would be super handy and nice to have it would diminish the value of previous dweets that are impressive exactly because they managed to get the characters for the hsl function. Furthermore, as this is a change to the "dwitter language", it might involve side effects, dweets aren't exactly clean code.

So why do we have R(), S() etc? They were there from the beginning before it was decided to freeze the language.

The last point is that more shortcuts moves dweets even further away from js, and the only thing you gain is more complexity which is effectively just more characters for certain dweets. (no-one used to argue that we should increase the length of dweets, but now there are of course 280 requests ^^)

For more discussion please see 
https://github.com/lionleaf/dwitter/issues/95 .  Feel free to continue the discussion there.

I'm closing this for now, but should there ever be a dwitter-lang 2.0 "better utility functions 3000"  this would be in :)

PR #291, opened 2018-02-14
Escape comment text to prevent simple script injection

AnnanFay:
This should fix the issue around simple script injection into dweet comments. I've not tested this. As it is right now you can simply enter `<script>anything</script>`.


lionleaf:
Oh wow, you're right :O  I think it must have snuck back in at some point, pretty sure it was properly escaped.

However, this fix doesn't seem to fix it when the comments are loaded from the template. And now that we are injecting a bunch of actual html in the comments with code blocks and links it's not as straight forward as just escaping it, or you'll lose the links/code blocks I suspect.

I'll have a closer look when I have time

lionleaf:
Solved by #321 and #320 :D

PR #296, opened 2018-03-02
Move user menu back to working position

moonrisewarrior:
Put the right margin back to default to fix misalignments on smaller screens.


moonrisewarrior:
Although I do think that ```margin-right: 0.7em;``` would look better in both themes. What do you think @lionleaf ?

lionleaf:
fixes #296

PR #301, opened 2018-03-22
Stop menu from being clickable when invisible

joeytwiddle:
When the profile dropdown menu is hidden, it can still be clicked. Try it!

This is especially an issue for low-res displays, where the invisible menu takes up a large proportion of the screen, so it can easily be tapped accidentally.

We could have used `display: none` to hide the menu. But this would prevent the `opacity` transition from being seen.

So instead I used `pointer-events: none` to make the menu un-clickable when it is hidden/collapsed.

(Actually the opacity transition is so fast I can hardly notice it. But since a previous developer put it there, I tried to keep it.)

*I have not tested this code!* Please comment if it works for you, on PC or on mobile.

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -395,6 +395,7 @@ iframe {
   right: 0.7em;
   z-index: 10;
   transition: opacity 0.05s ease-in-out, margin 0.05s step-end;
+  display: none;
   opacity: 0;
   height: 0;
   margin-top: 7px !important;
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -395,8 +395,8 @@ iframe {
   right: 0.7em;
   z-index: 10;
   transition: opacity 0.05s ease-in-out, margin 0.05s step-end;
-  display: none;
   opacity: 0;
+  pointer-events: none;
   height: 0;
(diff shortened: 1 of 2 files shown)


joeytwiddle:
Here is my blurb from the Discord:

`.collapsible` uses `opacity: 0` to hide itself. That's why the menu is still clickable even though it is invisible. :boom:

I guess we didn't notice that problem before because the header bar used to scroll off the top, so it only affected the first dweet.

The normal solution is to use `display: none` to hide the menu, or keep using opacity but add `pointer-events: none` to let clicks pass through. (`height: 0` on the `<ul>` might also work.)

Unfortunately those solutions, whilst fixing mobile, will break desktop, which relies on hover behaviour to open the menu!

That hover behaviour is somewhat problematic in itself, because on displays below 680px the menu overlaps the dweet, making it impossible to click the top-right of the dweet.

**My recommendation would be to completely remove menu's display-on-hover behaviour (often considered bad UX) and instead require an explicit click/tap to open the menu, on all devices.**

But now I am stuck. I can see how the hover works on desktop, but I don't know how the click on the profile avatar causes the menu to drop-down on mobile. What code does that?

PR #304, opened 2018-03-26
Some small mobile fixes + theme refactor

stianjensen:
After:
<pre align=center></pre>
<pre align=center></pre>

Before:
<pre align=center></pre>
<pre align=center></pre>


dwitterer:
Base.html - line 35 - the comment contains an old filename. Should be changed for clarity... But that's just my OCD kicking in...

Also, theme-dark.css and main.css sound odd... How about light-theme.css and dark-theme.css?

dwitter/templates/base.html
```diff
@@ -31,34 +31,34 @@
...
-	    l.setAttribute('href', '{% static "main_dark.css" %}');
-	  }
-	  else{
-	    l.setAttribute('href', '{% static "main.css" %}');
-	  }
-      }
-      else {
-        alert('Your browser seems not to support localStorage/cookies, therefore this feature will not work for you.');
+        var switchTheme = function(){
+          // Change the imported stylesheet from main.css to main_dark.css or vice versa
```
lionleaf: As noted by @dwitterer, this commennt now contains an old filename

stianjensen: Oops!

stianjensen:
> Also, theme-dark.css and main.css sound odd... How about light-theme.css and dark-theme.css?

I agree, but right now the one fila is the base from which the other builds on. So calling it light-theme wouldn’t be accurate either. I forst thought of having three files, one base and two themes, but it was easoer and better for the initial css loading to ha e the default theme inline.

PR #306, opened 2018-03-31
Add overflow hidden to canvas wrappers

### LINKED ISSUE #156: Iframes bleed over dweet cards while loading on iphone
Is this common iPhone behavior?

img_2017-02-28 23 26 49

stianjensen:
This fixes one long-standing and mega-annoying issue on at least iPhone
for *all* dweets that override the width of the canvas (a common
technique).

Can we all take a moment to appreciate the fact that it took me two
years to get around to fixing this? 

As @rikard-io noted in #156 it was just a matter of hiding the overflow. This works because the wrapper div around the iframe has already hard-coded the aspect ratio to 16:9 using the classic `padding-bottom: 56.25%` trick.

This fixes #156 

Before:
<pre align=center></pre>

After:
<pre align=center></pre>

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -320,6 +320,7 @@ input.comment-submit:active {
   position: absolute;
   width:100%;
   top: 0; bottom: 0; left: 0; right: 0;
+  overflow: hidden;
 }
 iframe {
   width:100%;


dwitter/static/main.css
```diff
@@ -320,6 +320,7 @@ input.comment-submit:active {
   position: absolute;
   width:100%;
   top: 0; bottom: 0; left: 0; right: 0;
+  overflow: hidden;
```
lionleaf: No wonder it took two years!

PR #309, opened 2018-03-31
Merge sharing options into one popover

stianjensen:
This creates a new, slightly nicer looking, popover that contains both
the share URL and an embed code. The plan is to re-use this popover
style for more types of content on the dweet card.

  
screenshot 2018-03-31 18 12 16

### DIFF
--- a/dwitter/static/js/feed.js
+++ b/dwitter/static/js/feed.js
@@ -31,9 +31,8 @@ $(document).ready(function() {
     var link = $(el).find('.fullscreen-button');
     var iframe = $(el).find('iframe.dweetiframe');
     var sharebutt = $(el).find('.share-button');
+    var shareContainer = $(el).find('.share-container');
     var sharelink = $(el).find('.share-link');
-    var embedbutt = $(el).find('.embed-button');
-    var embedsrc = $(el).find('.embed-src');
 
     link.on('click', function(e) {
       e.preventDefault();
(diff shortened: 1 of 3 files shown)


joeytwiddle:
I have one concern. Currently, when we click the 'share' link, the URL is immediately selected, so I can just hit Ctrl-C to copy. I find that quite convenient.

But it looks like you have removed that behaviour. Is there some reason you don't want to keep it?

stianjensen:
> I have one concern. Currently, when we click the 'share' link, the URL is immediately selected, so I can just hit Ctrl-C to copy. I find that quite convenient.

> But it looks like you have removed that behaviour. Is there some reason you don't want to keep it?

I intended to keep that behavior for the share link (while having to lose it for the embed link). If it's not auto-selected anymore, I'll look into it and try and fix it.

joeytwiddle:
Oops! I will make sure to get a local copy running before reviewing again. 🙇

lionleaf:
See #311 .  I guess I could've just checked out the last commit on prod, but I wanted to test the github revert feature :P 

Looks like you can't reopen a merged and reverted pull request. Will open it as a new pull request

lionleaf:
Alright, moved to #312 :P 

PS: No worries @joeytwiddle it's on me. Still doing some risky merges from time to time, but it's usually quick to revert anyway. I wanted to play with github revert features though.
Also; checking that stuff works on dweets that have been "scrolled in" seems to be quite important

joeytwiddle:
Yeah it was harmless enough, no state changes.

And I even got a brief look at the see-through links issue! Perhaps the share popover needs a bit of `z-index` to rise above them.

PR #312, opened 2018-04-06
Revert "Revert "Merge sharing options into one popover""

lionleaf:
Reverts lionleaf/dwitter#311

And hence reintroduces #309  :P Sorry for the clunky way of doing the pull requests. Fun playing with these features.


stianjensen:
I can have another go at this over the weekend. Didn't realize scrolled-in dweets were handled differently. If they use the same template it should only be a matter of changing the way I bind some event listeners.

stianjensen:
Replacing this PR with #313 since it's not strictly a simple revert anymore.

PR #313, opened 2018-04-08
Merge sharing options into popover again

### LINKED ISSUE #310: Visual bug on share card
With the new share box from #309 , the u/ and d/ seems to shine through the box.

Windows 10, Chrome

image

image

@stianjensen  ^

stianjensen:
This creates a new, slightly nicer looking, popover that contains both
the share URL and an embed code. The plan is to re-use this popover
style for more types of content on the dweet card.

When I previously did this in f73d5b0 I was not aware that the code used
for lazy-loaded dweets was completely separate in another js file.
So this time, I delete the duplicated code in scrolling.js and instead
implement it one way by binding event listeners once on body instead of
separately per dweet card. This is also better for performance when
having lots of dweets loaded.

screenshot 2018-03-31 18 12 16

This fixes #310

### DIFF
--- a/dwitter/static/js/feed.js
+++ b/dwitter/static/js/feed.js
@@ -27,47 +27,27 @@ $(document).ready(function() {
   $('body').on('input', '.code-input', onDweetChanged);
   $('.code-input').each(onDweetChanged);
 
-  $('.dweet').each(function(i, el) {
-    var link = $(el).find('.fullscreen-button');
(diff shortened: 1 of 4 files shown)


stianjensen:
Oh, and this also fixes a bug where the timestamps are formatted in UTC when they are first lazy loaded, and the timezone only corrected after you click the share link 😎

PR #314, opened 2018-04-08
Add popover with list of remixes

### LINKED ISSUE #302: Add "Remixed in: <list of dweet ids>"
Remixes link to the parent dweet, but I don't see the link from the parent to the remixes.

stianjensen:
This creates a more generic system for popovers and re-uses it for
listing remixes of a dweet.

This fixes #302.

<pre align=center></pre>

Depends on #313

### DIFF
--- a/dwitter/static/js/feed.js
+++ b/dwitter/static/js/feed.js
@@ -34,21 +34,26 @@ $(document).ready(function() {
     requestFullscreen(iframe[0]);
   });
 
-  $('body').on('click', '.share-button', function(e) {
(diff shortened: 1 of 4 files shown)


stianjensen:
I have no idea how this will work for dweets with lots and lots of remixes, but I also don't know how many remixes is the maximum a dweet currently has. I guess we'll find out soon enough!

lionleaf:
For many remixes maybe a maximum height and scroll for overflow?

Also, as mentioned by p01 at #302 maybe change the link text to "2 remixes" etc?

PR #315, opened 2018-04-09
First implementation of hashtags

lionleaf:
MVP for hashtags, taking feedback :D 

(Due a refactor of the view code; but I'll do that after the hashtags)

### DIFF
--- a/dwitter/feed/urls.py
+++ b/dwitter/feed/urls.py
@@ -39,5 +39,9 @@ urlpatterns = [
     url(r'^e/(?P<dweet_id>\d+)$',
         views.dweet_embed, name='dweet_embed'),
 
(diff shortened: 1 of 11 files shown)


lionleaf:
Don't have much more time today, will fix linting later :) Also want to add tests before merging.

lionleaf:
Important discussion to be had before merging this PR: What should constitute a legal #hashtag? 

At the moment it's any combination of [_a-zA-Z0-9] .  So #under_score is legal #hy-phen is not (#hy would become a hashtag). Starting with a number is legal. Starting with #_underscore is legal. It's case insensitive #SomeHashtag = #SOMEHASHTAG = #somehashtag   (but the case will follow to the url)

lionleaf:
Now with some basic tests as well! I'm ready for a review and potential deploy :D

lionleaf:
Merging on verbal review from @sigvef

PR #318, opened 2018-04-09
Restore the href on the share-button

### LINKED ISSUE #317: All the 'share' links have an empty href
In `dweet_card.html` we used to have:

    <a class="dweet-option share-button" href="{% url 'dweet_show' dweet_id=dweet.id %}" target="_blank">share</a>

but now we have:

    <a class="dweet-option share-button" href="" data-popover=".share-container">share</a>

That means we cannot right-click on the share link to open that dweet in a new tab, which is something I find useful. Is there another way I should be doing that?

I will try to restore it.

(Full disclosure: I originally put that href there back in c6dc918210a474e73b5f86f2a8ad4e4d907bb88f)
(It was removed in @stianjensen's recent 7f8a782b94ac8397fff48d725fad54e85f0bcd29, possibly for reasons)

joeytwiddle:
Closes #317 

I see Stian is still using `e.preventDefault()` in https://github.com/lionleaf/dwitter/pull/314, so I hope this href won't break the popover. But it will be about 8 hours before I can get home and test that...!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ Inspired by [arkt.is/t/](http://arkt.is/t/Yy53aWR0aD0yZTM7eC5maWxsUmVjdCgxNTAsMT
 * `make` (install dependencies and set up database)
 * `python manage.py createsuperuser` (create admin account used below)
 * `make run` runs the server. Use `make serve` instead if you're working inside a VM with port forwarding. (0.0.0.0:8000)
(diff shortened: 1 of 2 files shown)


stianjensen:
Sorry for removing that href, what is it used for?

joeytwiddle:
No worries Stian! Perhaps I should leave a comment for the non-obvious code.

I have tested this on local and it's working as expected. (I added the `target="_blank"` back too.)

PR #324, opened 2018-04-14
Give dweet pages titles (either dweet ID or author's comment)

joeytwiddle:
### What it do?

It gives semantic titles to the individual dweet pages. This will be good for bookmarking, and for search engine indexing.

### Requests for assistance

1. Do we need to escape the comment string before we put it in the HTML, or does Django do that for us?
2. ~I wanted to truncate the title to a maximum length of 40, but couldn't figure out the Django syntax (see comment in code)~
3. ***Done:*** Or we could simply ignore comments which are longer than 40 chars, and assume they aren't title-worthy.
4. The nested if-blocks barf a lot of spaces into the rendered HTML output. Perhaps this new titling code should move into a Python function somewhere.
5. (Future) Do you want to add "(138 chars)" at the end of the title?
6. (Future) If the dweet author has not left a comment, should we take the first comment and use that? ([example](https://www.dwitter.net/d/6473))
7. (Future) We could add titles for some of the other routes too, e.g. "Top feed - Dwitter" and "magna's awesomed dweets - Dwitter"

### Screenshots (check the window titles)

If the dweet author has made a comment which is 40 chars or less, that comment becomes the page title:

joey s desktop on tomato at 7 04 52 pm on saturday 14 april 2018

Otherwise the dweet ID is used in the page title:

joey s desktop on tomato at 7 05 04 pm on saturday 14 april 2018

Thanks to prplz for some guidance and improvements

### DIFF
--- a/dwitter/templates/base.html
+++ b/dwitter/templates/base.html
@@ -2,7 +2,21 @@
 <html>
   <head>
     <meta charset="utf-8" />
-    <title>{% block title %} Dwitter {% endblock %} </title>
+    <title> {% block title %}
+      {% if dweet %}
+        {% if dweet.comments.last.author == dweet.author %}
+          {% with dweet.comments.last.text as author_comment %}
+            {# TODO: Djangify this: (author_comment[:37] + '...') if len(author_comment) > 40 else author_comment #}
+            {{ author_comment }}
+          {% endwith %}
+        {% else %}
+          d/{{ dweet.id }}
+        {% endif %}
+        by {{ dweet.author.username }} - Dwitter
+      {% else %}
+        Dwitter
+      {% endif %}
+    {% endblock %} </title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     {% load compress %}
 
--- a/dwitter/templates/base.html
+++ b/dwitter/templates/base.html
@@ -2,21 +2,7 @@
 <html>
   <head>
     <meta charset="utf-8" />
-    <title> {% block title %}
-      {% if dweet %}
-        {% if dweet.comments.last.author == dweet.author %}
-          {% with dweet.comments.last.text as author_comment %}
-            {# TODO: Djangify this: (author_comment[:37] + '...') if len(author_comment) > 40 else author_comment #}
-            {{ author_comment }}
-          {% endwith %}
-        {% else %}
-          d/{{ dweet.id }}
-        {% endif %}
-        by {{ dweet.author.username }} - Dwitter
-      {% else %}
-        Dwitter
-      {% endif %}
-    {% endblock %} </title>
+    <title>{% block title %} Dwitter {% endblock %} </title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     {% load compress %}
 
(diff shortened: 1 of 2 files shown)


lionleaf:
See inline for some opinions on your points:

### Requests for assistance

1. _Do we need to escape the comment string before we put it in the HTML, or does Django do that for us?_
Should be sorted:
screen shot 2018-04-16 at 7 17 07 pm

2. ~I wanted to truncate the title to a maximum length of 40, but couldn't figure out the Django syntax (see comment in code)~
**If you want to truncate I think you could do:**
`{{dweet.comments.last.text | truncatechars:40}}`

4. _The nested if-blocks barf a lot of spaces into the rendered HTML output. Perhaps this new titling code should move into a Python function somewhere._

**Eh, whatever**

5. (Future) Do you want to add "(138 chars)" at the end of the title?
**Yes, that would be cool**
6. (Future) If the dweet author has not left a comment, should we take the first comment and use that? ([example](https://www.dwitter.net/d/6473))
**No, I don't think we should let other people define it. A better solution is to add a text field when you submit the dweet for the first comment, which should have people using it a whole lot more**

7. (Future) We could add titles for some of the other routes too, e.g. "Top feed - Dwitter" and "magna's awesomed dweets - Dwitter"

**Yes**


Other points:
- As titles get truncated in browsers, order matter a lot. Maybe something like "Dwitter - lionleaf - "My comment" (122/140) - d/123"?
- Do we want the ID to always be in there?

joeytwiddle:
1, 3, 5, 6, 7. Cheers for feedbacks!

2. I decided against truncating in the end. I assume if the comment is really long, it is more likely to be a conversation than description of the dweet, which won't look good in bookmarks.

> As titles get truncated in browsers, order matter a lot. Maybe something like "Dwitter - lionleaf - "My comment" (122/140) - d/123"?

I see your point, but disagree. I think the _specific_ is usually what we care about. I'm looking through my bookmarks for that cool "lava lamp", without necessarily remembering which site it was on.

I like it when other sites order titles by reverse-significance too. For example:

https://medium.com/@Creativesoulw/the-banana-exercise-1f9991c9dfbc
The Banana Exercise – Creative Soul Writing – Medium

> Do we want the ID to always be in there?

I did consider that, for consistency. But in the end I think we don't care that much about the ID. I used it more as a placeholder when no title is available.

Short is beautiful. Less is more. (We technical types sometimes forget that.)

lionleaf:
2. I think maybe truncation is better. For one I don't like the non-obviousness of an arbitrary length limit on that feature. Secondly I think there are many cases where the truncated comment would make just as much sense, for instance from the front page now: https://www.dwitter.net/d/7693


As for the ordering of the titles, I think we use titles differently. My main use case is to spot which among my 100 tabs is what website. However, once we get a favicon that is solved. So, point taken. I'm happy with your ordering

joeytwiddle:
12 hours later, tried and tested. :)

That's a nice Django function. I notice the (pipe-to-a-filter) syntax is quite similar to Angular's. I wonder who stole it first?

We could possibly extend the arbitrary cap from arbitrary 40 to arbitrary 50, but keeping it short for now leave room for us to add more noise in the future, and really dwitterators should golf their titles anyway!

Issue #279, opened 2018-02-03
Login and registration doesn't show errors


lionleaf:
Errors like "wrong username/password", "that user already exists" etc etc are not being shown on registration and login page.

Should also add a note on the login page that usernames are used, not emails. And that it's case sensitive

diceroll123:
I'm kinda conflicted here, you can have two users, `steve` and `STEVE`... 

Well, the development build lets me do that, I'm unsure if prod is doing it. 🤔 Is this intentional?

lionleaf:
It's default behaviour from djang-registration-redux; so it's the current behaviour. Enforcing a new policy on that going forward is probably a good idea :)

However, that's a separate topic. This issue is just the lack of visual feedback when things are wrong. Right now it just doesn't respond.

stianjensen:
Case sensitive user names is the default in Django, even.

dwitterer:
Fixed in #325. You should close this issue or merge it with that pull request or something...

Issue #177, opened 2017-03-09
Feature request: share dweet as gif


snnw:
The rendering should probably be done client-side. Something like https://github.com/jnordberg/gif.js/ should do the trick. Uploading to twitpic through their API and integration with Twitter would be great.

iver56:
That would be awesome 💯 Feel free to create a pull request 👍

aleksanb:
If you use MediaRecorder you can record directly to a webm from a canvas, the same way that the site https://www.musicvideodispenser.com/ does it.
Then the webm can be shared (and twitpic can reencode it to a gif if they want!)

Currently there are some implementation issues to make such a thing for dwitter as the canvas is within an iframe, so the recording would have to be inside the content loaded in the iframe, which breaks separation somewhat.

mxfh:
would welcome this, doesn't even need another website or api, but export button would be easy, just pull each frames toDataURL and let some workers pack them into a gif, as already mentioned see: https://jnordberg.github.io/gif.js/

lionleaf:
Yeah, this is a great idea we're coming back to from time to time. If it can be done client side that would be awesome, as I'm a tad bit resistant to add too much server side computation. Very happy to have survived hackernews front page on the smallest digitalocean droplet without issues.

Doing it client side also means you're sharing what you're seeing, in the cases where it might be different on the server (emoji versions for instance)

iver56:
Maybe the client side rendered gif could also be stored on the server, so it can later be referred to in open graph ("og") meta tags, so the dweet looks nice when its link is shared on social media

atesgoral:
Could be useful for this: https://developers.google.com/web/updates/2017/04/headless-chrome

lionleaf:
Implemented in #325

PR #325, opened 2018-04-14
Issue fix and record functionality

BalintCsala:
- Fixes issue #279 (errors on invalid login credentials)
- Adds recording functionality (GIF)


dwitter/templates/dweet/dweet.html
```diff
@@ -7,10 +7,20 @@
...
       var autoplay = getURLParameter('autoplay');
       var lastError = "";
       var playing = !!autoplay;
+    var FPS = 60;
+    
+    var recording = false;
+    var encoder = new GIF({ 
+    workers: 4, 
+    quality: 5, 
+    workerScript: "http://dweet.localhost:8000/static/libs/gif.worker.js"
```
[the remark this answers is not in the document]
prplz: https://docs.djangoproject.com/en/1.8/howto/static-files/ :)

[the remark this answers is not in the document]
joeytwiddle: Will this work? `workerScript: "/static/libs/gif.worker.js"`

dwitter/static/js/scrolling.js
```diff
@@ -16,6 +17,18 @@ window.onload = function() {
     registerStatsClickListeners(dweet);
   });
 
+  [].forEach.call(recordButtons, function(recordButton) {
```
lionleaf: This only attaches the event to the first page of dweets, and is not called when new dweets are loaded with ajax when you scroll to the bottom. (Yeah, I know it should be done differently :P )

lionleaf: Take a look at how stiaje did it for the remix link pull request

dwitter/templates/dweet/dweet.html
```diff
@@ -7,10 +7,18 @@
...
+  <script type="text/javascript" src="/static/libs/gif.js"></script>
     <script>
       var autoplay = getURLParameter('autoplay');
       var lastError = "";
       var playing = !!autoplay;
+    var FPS = 60;
+    
+    var recording = false;
+    var encoder;
+	var gifCanvas = document.createElement("canvas");
```
lionleaf: Just can't help pointing it out: You're mixing spaces and tabs here.

dwitter/templates/dweet/dweet.html
```diff
@@ -39,6 +47,38 @@
           case "hideStats":
             setStatsVisibility(false);
             break;
+          case "record":
+            recording = true;
```
lionleaf: Check if recording is already true here? It might be possible to send two "record" messages in a row if you click fast?

dwitter/templates/dweet/dweet.html
```diff
@@ -39,6 +47,38 @@
...
+            encoder = new GIF({ 
+              workers: 4,
```
lionleaf: Could be worthwhile to change this depending on hardware? I'm not sure how well supported it is; but found this: https://stackoverflow.com/questions/3289465/get-number-of-cpu-cores-in-javascript?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa 

Should have a default fallback if the API isn't available.

Not critical, so I'm happy to merge this with `workers: 4` and open an issue instead

dwitter/templates/dweet/dweet.html
```diff
@@ -39,6 +47,38 @@
...
+              height: c.width / 16 * 9,
+			  background: ""
+            });
+            FPS = 25;
+			gifCanvas.width = c.width;
+			gifCanvas.height = c.width / 16 * 9;
+			gifctx = gifCanvas.getContext("2d");
+            break;
+          case "stop":
+            recording = false;
```
lionleaf: Same as above: Check that we are actually recording before stopping one

dwitter/templates/dweet/dweet.html
```diff
@@ -53,7 +93,7 @@
           parent.postMessage({
             'type': 'error',
             'error': ""+e,
-            'location': window.location.href
+            'location': location.href
```
lionleaf: Why was this change made?

dwitter/templates/dweet/dweet.html
```diff
@@ -144,15 +185,21 @@
       if (playing){
         requestAnimationFrame(loop);
       }
-      time = frame/60;
-      if(time * 60 | 0 == frame - 1){
+      time = frame/FPS;
```
lionleaf: As mentioned earlier; for some dweets, changing this when recording a gif will make the dweet behave differently. Which is not ideal. However, I don't think too many dweets are affected.

I say we do one of two things:
1. Say that this is fine for now, and if there are issues with some dweets we will look into it later
2. If you change the gif FPS to 30 or 15 you could do `if(recording && frame % 2 == 0)` to capture every other frame and still run the dweet exactly as expected.  (15 might be enough?)

What do you think?

dwitter/templates/registration/login.html
```diff
@@ -17,7 +17,7 @@
         type="text"
         placeholder="..."
         />
-
+  <span class="error-display">{{ form.username.errors }}</span>
```
lionleaf: Nice, thanks!

dwitter/static/js/scrolling.js
joeysino: Here's the magic: https://github.com/lionleaf/dwitter/pull/313/files#diff-166af2ad95d6cae2973ae18decaa1016R37

This asks jQuery to listen for _all_ click events on the body, but it filters them by the selector, so it only actually calls the handler function if the click was made on or below a `.share-button` element.

Disadvantages: Any click anywhere will have to be checked by jQuery. (Slightly more CPU)
Advantages: We only have to add one event listener to handle all the record buttons, even buttons added in the future. (That uses less memory than a separate event handler for each individual record button)

BalintCsala:
I'm no gif wiz either, 60 frames would have been too much, 10-15 not enough, 24 (cinema quality) would have created fractions and I don't think it can handle those, so I went with 25.

dwitter/templates/dweet/dweet.html
BalintCsala: Ah, sorry, I thought I replaced them.

dwitter/templates/dweet/dweet.html
BalintCsala: It can only send a stop message after you press record. The message depends on wheter the record button has the class name "recording" and thag's the first thing I change

dwitter/templates/dweet/dweet.html
BalintCsala: I think I was just confused at one point and checked whether the window still contained the same things as the global scope, then forgot to change it back.

dwitter/templates/dweet/dweet.html
lionleaf: Fair enough. In this case because you do the toggle of the class directly in the click event I don't see way you will get anything but every other 'record' and 'stop', as js is single-threaded in this case.

I was assuming there was a bounce to the actual dweet confirming that it is indeed recording before switching. I guess this means it will tell you it is recording even if it went wrong :P I think it's fine as it is then.

dwitter/templates/dweet/dweet.html
```diff
@@ -39,6 +47,38 @@
           case "hideStats":
             setStatsVisibility(false);
             break;
+          case "record":
```
lionleaf: rename to startGifRecording or something similar. (camelCase to be consistent with hideStats, I guess)

dwitter/templates/dweet/dweet.html
```diff
@@ -39,6 +47,38 @@
...
+              width: c.width, 
+              height: c.width / 16 * 9,
+			  background: ""
+            });
+            FPS = 25;
+			gifCanvas.width = c.width;
+			gifCanvas.height = c.width / 16 * 9;
+			gifctx = gifCanvas.getContext("2d");
+            break;
+          case "stop":
```
lionleaf: Rename to `stopGifRecord` as "stop" sounds like you're stopping the dweet, or anything else.

lionleaf:
fixes #177 
fixes #279

PR #337, opened 2018-04-29
Slight tweaks to favicon and code blacks in dark theme

dwitterer:
I made the code blocks (the ones in the comments) look much nicer on dark theme, and made it a teeny tiny (~2%) bit lighter on the main theme, because the values seemed a bit weird. I think it looks very nice now :D 

Also made the favicon have a white background so it works better on mobile/Firefox Quantum etc...

A side-by-side of the new colors is provided, as well as the current colours:

NEW:
yayornay

CURRENT:
current


dwitter/static/main.css
```diff
@@ -207,8 +207,8 @@ ul.messages li.success{
   font-size: 13px;
   line-height: 25.28px;
   padding: .25em;
-  background: hsl(250, 56%, 95%);
-  color: hsl(250, 56%, 53%);
+  background: hsl(0, 0%, 92.5%);
```
lionleaf: Did you mean to remove the purple tint of the background on the light theme?

dwitterer:
Yeah... it's still kind of present though, and it the end I only made it lighter...

dwitter/static/main.css
dwitterer: Yeah... it's still kind of present though, and it the end I only made it lighter...

PR #340, opened 2018-04-29
Update favicon

### LINKED ISSUE #339: Fix the width of the bars on the favicon
During recent changes, me and lionleaf have somehow made it so that the bars on the favicon are very wide. There are different options for solving this, and so I'm opening this issue for people to discuss and suggest different configurations for the bars. Generally, lionleaf wants:

- The height of the bars to be 4x their width
- The gaps between the bars to be same as their width
- And ideally, the margins on either side of the bars to be the same size

dwitterer:
I had a little HTML test, and I think this should fix the favicon issue for now... I has an extra 1 px on the right but it still looks much better than the current icon in terms of the bars.

This PR could close #339


lionleaf:
This has been hanging around for a long time, but I believe the aspect ratio on the boxes was the reason it's not been merged. Closing for now, and we can reopen it later if need be.

PR #348, opened 2018-06-11
Fix remixes not enforcing 140 char rule

PythEch:
As a matter of fact I was able to bypass the 140 characters limit. This is because reply/remix API does not enforce the 140 character limit rule, unlike the one responsible for creating dweets from scratch. This was fixed by copying and pasting the preceding code.

PoC: https://www.dwitter.net/d/8555 (no hard feelings if deleted)

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -205,8 +205,13 @@ def dweet(request):
 
 @login_required
 def dweet_reply(request, dweet_id):
+    code = request.POST['code']
+
+    if(len(code.replace('\r\n', '\n')) > 140):
+        return HttpResponseBadRequest("Dweet code too long! Code: " + code)
+
     reply_to = get_object_or_404(Dweet, id=dweet_id)
-    d = Dweet(code=request.POST['code'],
+    d = Dweet(code=code,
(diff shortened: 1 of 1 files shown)


PythEch:
Also, looking at this commit: d527e431c4c359287ba2e7a32900db27afea7de6

It seems inconsistent on the front-end side, as currently "Post as remix" button is disabled after 140 chars but "Post dweet!" button is not.

Handling this possibility is a must on the back-end side though if we want to disallow such exploits. But maybe front-end also could be edited to be consistent.

lionleaf:
Thanks! I'll get this merged once I have time to verify it locally, but looks good :)

Good point on the "Post dweet!" not being disabled when you go above 140 characters. Should be added at some point (feel free to raise an issue on that).

And, yes, we've had a number of discussions on how to count characters.  Consistency is key, so if there are inconsistencies on the frontend that is consistent between browsers I'd like to get that fixed somehow. It's a fiddly one

PythEch:
I can confirm the exploit no longer works. About inconsistency, I am thinking about sending a pull request when I have free time.

PR #350, opened 2018-06-17
Adds polyfills

BalintCsala:
This is aims to fix the different compatibility issues across browsers (mostly IE and egde). It solves the issue, where Edge doesn't clear the path and in some cases the background, when c.width is modified and replaces some Math function IE lacks (cosh, cbrt, trunc, fround, etc.)

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -87,11 +87,11 @@
             var width = Math.max(c.width, minGIFWidth);
             if (typeof navigator.hardwareConcurrency !== "undefined")
               workerCount = navigator.hardwareConcurrency;
-            encoder = new GIF({ 
-              workers: workerCount, 
-              quality: 5, 
+            encoder = new GIF({
+              workers: workerCount,
+              quality: 5,
               workerScript: "/static/libs/gif.worker.js",
-              width: width, 
+              width: width,
               height: width / 16 * 9,
               background: ""
             });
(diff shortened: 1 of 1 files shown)


dwitter/templates/dweet/dweet.html
```diff
@@ -277,31 +278,10 @@
...
-          c._height = value;
-        },
-        get: function() {
-          return c._height;
-        },
-        configurable: true
+      $(c).resize(function() {
+        x.beginPath();
+        x.resetTransform();
+        x.clearRect(0, 0, obj.width, obj.height);
```
joeytwiddle: So `'resize'` gets triggered even if the canvas didn't change size? wtfdom!

lionleaf:
Great! On a Mac atm, do most dweets work correctly in IE and Edge after this? And any idea of potential performance implications?

dwitter/templates/dweet/dweet.html
joeysino: Mmm yeah, regarding @lionleaf's performance question, will this `clearRect` run on all platforms, whenever `c.width` is assigned?

Won't that impact performance on platforms where it's not needed?

BalintCsala:
@lionleaf It doesn't seem like it has any performance effects, the dweets have the same FPS on my local version and on dwitter.net

BalintCsala:
Also, yes, this should solve most of the problems with edge and if we notice something else not being reset, then we can add that too easily

Issue #331, opened 2018-04-22
Cancel gif processing


lionleaf:
Gif processing takes quite a long time, especially if it let it record for >5s. Feature request: press the 'recording' button to cancel the recording

lionleaf:
Closed by #352 Thanks @PythEc

lionleaf:
Closed by #352 Thanks @PythEc

PR #352, opened 2018-06-24
Added ability to stop GIF processing, handling #331

PythEch:
Uses gif.js’s undocumented 'abort()' function that stops the workers


PythEch:
Good idea! If possible, we can also add a progress indicator.

Issue #66, opened 2016-08-19
Completely unload dweets that are outside the screen


lionleaf:
Even though dweets stop running, they still fill up memory, after scrolling through a couple I easily hit 1GB memory usage on a dwitter tab. Solution would be to free/unload dweets after they have been scrolled away from (say 2 dweets out of screen)

PythEch:
Somehow after scrolling a lot I also get huge frame drops on my MacBook Pro with integrated GPU in addition to the memory leak you mention. On my desktop computer however I don't have this problem which has a relatively powerful GPU. Performance profiler indicated that the problem was due to extensive use of GPU.

lionleaf:
That's interesting. Some kind of HW acceleration of the canvas must be at play. And you're sure it isn't just the active dweet? Full destruction of the iframe is probably the best way to go.

PythEch:
It isn't just the active tweet. The framedrops affect the `default`. Normally I get 60 FPS which shouldn't be an issue, I mean it's the default, right? But after scrolling down a bit, 10-20 dweets maybe, and then scrolling up to the top, I notice I have only 30 FPS on the default dweet.

The behaviour is consistent with Chrome, Chromium and Opera which all use Blink engine. Much surprisingly however, Safari doesn't have this issue, along with Firefox but both have worse rendering performance than Chrome.

Scrolling also is jaggy on Blink browsers, unlike Firefox and Safari.

PR #355, opened 2018-06-25
Stops infinite loops

BalintCsala:


joeytwiddle:
Great stuff Balint!

Does the loopbuster run on all dweets, or only on dweets which are being edited?

Here are some dweets which make heavy use of recursion. They may be good for testing.

https://www.dwitter.net/d/2727 One-time cloud by joeytwiddle
https://www.dwitter.net/d/2737 Animated clouds by nystrand
https://www.dwitter.net/d/2738 Animated clouds 2 by nystrand
https://www.dwitter.net/d/7638 Lots of u() calls by joeytwiddle

I should also try to find one with a heavy loop inside it...

dwitter/static/libs/loopstopper.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+  },
+  testLoop: function(loopId) {
+    if (this.loops[loopId] == 3)
+      return true;
+    if (this.loops[loopId] == 1) {
+      if (Date.now() - this.loopTimers[loopId] > this.TIME_OUT_LENGTH) {
+        this.loops[loopId] = 3;
+        return true;
+        console.log(1);
```
lionleaf: dead code

dwitter/static/libs/loopstopper.js
```diff
@@ -0,0 +1,43 @@
...
+  // 1           - Currently being monitored
+  // 2           - Exited
+  // 3           - Infinite
+
+  reset: function() {
+    this.loops = [];
+    this.loopTimers = [];
+  },
+  restartLoop: function(loopId) {
+    if (!this.loops[loopId] || this.loops[loopId] == 2) {
```
lionleaf: So this should pass if it's 0 or 2. Is the main point to avoid reseting a 3 to a 1 or a resetting the timer on a loop in state 1?

dwitter/static/libs/loopstopper.js
```diff
@@ -0,0 +1,43 @@
+window.stopper = {
+
+  loops: [],
+  loopTimers: [],
+  TIME_OUT_LENGTH: 500,
+
+  // States:
```
lionleaf: Is there a significant performance penalty to use some sort of enum here to increase readability? Not a big deal as there isn't too much code, and it is quite performance critical

dwitter/static/libs/loopstopper.js
lionleaf: Having found the way instrumented code looks, I realize this is to make sure it's not in state 1.

Logic could be made more obvious here (maybe just a comment)

dwitter/templates/dweet/dweet.html
```diff
@@ -175,6 +174,10 @@
     {% load staticfiles %}
     {% compress js %}
     <script src="{% static "libs/stats.min.js" %}"></script>
+    <script src="{% static "libs/gif.js" %}"></script>
+    <script src="{% static "libs/loopbuster.js" %}"></script>
+    <script src="{% static "libs/loopstopper.js" %}"></script>
+    <script src="{% static "libs/jquery.min.js" %}"></script>
```
lionleaf: Hmm... Is jquery really required? This loads jquery for every single dweet and also extends the dweet language to have access to jquery functions

dwitter/templates/dweet/dweet.html
```diff
@@ -209,9 +212,15 @@
...
     var frame = 0;
+    
     function u(t) {
-       {{ code | safe }}
-      }
+      {{ code | safe }}
+    }
+    
+    
+    let src = (u + "").split(/[{}]/g)[1];
```
lionleaf: Haven't had the chance to play with this, but couldn't you craft a dweet that would break this?

lionleaf:
screen shot 2018-06-25 at 10 52 21 pm
I also seem to be running into some errors. That one is from https://www.dwitter.net/d/123 for instance . (try the code in your local dev environment)

dwitter/static/libs/loopbuster.js
```diff
@@ -0,0 +1,4235 @@
+(function() {
```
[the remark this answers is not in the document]
BalintCsala: That's from here: https://github.com/CodePen/InfiniteLoopBuster
I browserified it, so loopbuster is basically a compressed version of app.js and esprima.

dwitter/templates/dweet/dweet.html
BalintCsala: That one's for the resize polyfill, but I think I know of a way to remove it

BalintCsala:
@joeytwiddle It runs on every dweet and it solves recursion as well. Basically anything, that takes more, than 500 milliseconds

dwitter/static/libs/loopbuster.js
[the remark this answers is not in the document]
joeytwiddle: Could you add a comment with the command needed to do the browserifying?

In case someone needs to rebuild it again in future...

joeytwiddle:
Got it. Then those examples should work fine except the first one, the cloud, which takes a long time to produce the first frame. I don't have the URLs of any other dweets that do heavy processing like that, but there might be one or two. ***Share if you know any!***

I guess it makes sense to discourage people from making such heavy dweets anyway. (My cloud does lock up the browser whenever it gets loaded onto a page.)

But, perhaps it would make sense to increase the timeout, to 5 or even 10 seconds. (Like a fire alarm shouldn't go off when someone burnt the toast; it should only go off when it definitely detected a fire.)

I will try this feature out on my slow laptop later...

joeytwiddle:
Hmm I tried on the clouds (which are recursive), and I also took [this starfield](https://www.dwitter.net/d/7370) and gave it a much larger loop. But I didn't get any breakout.

I added a bit of logging to `loopstopper.js` and what I observed for both dweets was:
- `Date.now() - this.loopTimers[loopId]` was always 0 inside `testLoop()` (with status `MONITORED`)
- `restartLoop()` was being called more often than I would expect (with status `FINISHED`).

I'm afraid I can't dive any deeper right now.

BalintCsala:
@joeytwiddle Strange, it doesn't seem to work well with bodyless loops, like `while(1)doSomething()`

dwitter/templates/dweet/dweet.html
```diff
@@ -235,6 +234,11 @@
 
       try {
         stats.begin();
+        if ((u + "").match(/c\s*\.\s*width\s*(=|\+=|-=|\*=|\/=|%=|\**=|<<=|>>=|>>>=|&=|\^=|\|=)/) != null) {
+            x.beginPath();
+            x.resetTransform();
+            x.clearRect(0, 0, c.width, c.height);
+        }
```
PythEch: This wouldn't work on the very rare case of `c.height` (I've seen few dweets using it, but can't remember off the top of my head).
Besides, I think something like
```javascript
c.addEventListener("resize", function() { ... });
```
or
```javascript
c.onresize = function() { ... };
```
would be better I think, but I don't have any Windows machine to test this on at the moment.

BalintCsala: @PythEch  There's no resize event on anything other than the window. Fixing it shouldn't be hard, I only need to change `c\.width` to `(c\.width|c\.height)`

PythEch: You're right, totally forgot about that. Maybe adding a Proxy or defining a getter/setter would be better then. Or we could inspect how jQuery defines resize event on canvas. 

It just feels wrong IMO doing this for every frame for every other browser for performance reasons. 

On the correctness side, it looks like now you can just do:
```javascript
for(...)x.beginPath('c.width='),...
```
or
```javascript
//c.width=
```

now these two examples do not reduce the character count so they are not exploitable, debatable if it's worth the trouble but I could come up with at least one scenario where it reduces char count by 1:
```javascript
x.beginPath(x.fillStyle=`hsl(${i*t},99%,39%/*c.width=`)
```
compared to
```javascript
c.width|=0
x.beginPath(x.fillStyle=`hsl(${i*t},99%,39%`)

you could argue that such an advanced user could already do something like
```javascript
for(c.width|=i=9;i--;)...
```
to save 1 char but not every dweet is applicable for that.

BalintCsala:
@PythEch There you go, that should solve both problems, since it limits the check to Edge to make sure it doesn't affect performance and thus you can't really abuse it either if you target more, than Edge

PythEch:
Great stuff, love it!

My only complaint so far is that it has some performance penalty, about 5 FPS for a [47 FPS dweet.](https://www.dwitter.net/d/8774)

Maybe I'm [more than unlucky compared to average dwitter user](#66 ) 😄 but I'm in favor of having it enabled only on editing mode.

BalintCsala:
@PythEch That would defeat the purpose of this. It's mostly against harmful dweets

Issue #93, opened 2017-01-14
Completely clear canvas when dweet is changed including custom variables


lionleaf:

aemkei:
👍 This also made it hard for me.

joeytwiddle:
I tried deleting all the properties from `window`. But the browser didn't recreate some of them (e.g. `window.performance`) and so that idea died quickly.

I guess the idea of creating a whole new <iframe> is avoided because it would slow down the editing process, increasing the time between editing and seeing results.  But how about a compromise?  Load a new <iframe> every 40 edits, or every 1 minute, or when the mouse hovers over the Post button.  It won't catch the mistake immediately, but it should at least catch it before publishing.

atesgoral:
One option is to take a snapshot of all default globals and then selectively deleting only globals that weren't in that initial set.

lionleaf:
Solved

PR #358, opened 2018-07-01
Clear all newly added globals when dweet resets

joeytwiddle:
An implementation of [Magna's suggestion](https://github.com/lionleaf/dwitter/issues/93#issuecomment-401560322)

I gave it a little test (with the console log uncommented):

joey s google-chrome at 4 11 04 pm on sunday 1 july 2018

It seems to have detected a var `pauseTime` which we forgot to declare in the dwitter code. So I added a declaration for that too.

This code makes use of `Array.prototype.includes` which is [not supported in IE11](https://caniuse.com/#feat=array-includes). I could fix that if we want to support IE11.

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -19,6 +19,7 @@
       var lastError = "";
       var playing = !!autoplay;
       var FPS = 60;
+      var pauseTime = null;
       
       var recording = false;
       var encoder;
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -264,7 +264,9 @@
     function reset(){
       Object.keys(window).forEach(key => {
         if (!browserGlobals[key]) {
-          // console.log("Removing new global: " + key);
+          if (window.debug) {
+            console.log("Removing new global: " + key);
+          }
           delete window[key];
         }
       });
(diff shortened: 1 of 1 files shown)


lionleaf:
Great! Should get rid of most broken dweets. 
I'm guessing code editing still is equally responsive?

(I don't have my laptop at the moment, so will do a proper review / merge later)

Fixes #93

joeytwiddle:
Mmm yeah the code was a bit O(n^2)-ish.

I think it's easier for me to add a couple of lines to make it O(n) than to try to measure performance on a slow device. :stuck_out_tongue: Done!

dwitter/templates/dweet/dweet.html
```diff
@@ -181,7 +182,14 @@
...
     var stats = createStats();
     setStatsVisibility({{ newDweet }});
-  
+
+    // Remember the default globals so that later we can remove any that
+    // were created by the dweet
+    var browserGlobals = {};
+    Object.keys(window).forEach(key => {
+      browserGlobals[key] = true;
+    });
```
joeytwiddle: Some "functional programming" people recommend this style instead:
```js
var browserGlobals = Object.keys(window).reduce((a, key) => (a[key] = true, a), {});
```
or even:
```js
var browserGlobals = Object.keys(window).reduce((a, key) => ({...a, [key]: true}), {});
```
I could change to one of those, if you prefer. 😛

lionleaf: Fun way of doing it; but I'd say leave it as is now. Easier to read for a greater number of people.

PR #363, opened 2018-07-31
Green theme! + shift themes button update

kipkat:
## Green Theme
This Sci-Fi Hacky Lime green on Black background CSS makes the website look awesome. I am not able to test the new theme switch button and I don't know if it is working. 
image
I will quickly fix any bugs.

### DIFF
--- a/dwitter/static/theme_green.css
+++ b/dwitter/static/theme_green.css
@@ -1,0 +1,161 @@
+textarea {
+  border: none;
+}
+
+ul.messages li{
+  border:1px solid #0cff00;
+}
+
+.dweet{
+  background: #011900;
+}
+
+.content-div {
+  background:#000000;
+  color: #00ff47;
+  border:1px solid #0c5800;
+}
+
+.content-div input {
+  border:1px solid black;
+}
+.dweet, .card, .loading, .end-of-feed {
+  border:1px solid black;
+}
+
+.comment-input
+{
+  background:#3a0000
+}
+
+.submit-box {
+  border:1px solid black;
+}
+
+.dweet-create-form-title {
+  background: #00a708;
+}
+
+.dweet-create-form-title:hover {
+  text-decoration: underline;
+}
+
+.dweet-create-form-title .add-icon {
+  color: rgb(0, 78, 4);
+}
+
+.head-menu {
+  background-color: #4b4b4b;
+  border-bottom:1px solid black;
+}
+
+.head-account-info .collapsible {
+  border: 1px solid black !important;
+}
+
+.head-account-info.is-logged-in:hover {
+  background: #00ff18;
+}
+
+.head-account-info .collapsible li {
+  border-bottom: none;
+  color: #ff0606;
+}
+
+.header-title {
+  color: #00ff24;
+}
+
+.like-button {
+  border: 1px solid black;
+}
+
+.dark-section .dweet-code textarea {
+  border: none;
+}
+
+.dark-section .dweet-code textarea:focus {
+  border: none;
+}
+
+/* Dark-theme-specific properties */
+
+.head-menu a {
+  color: #17ff00;
+}
+
+.head-menu
+{
+  background: #141d14;
+}
+
+.body{
+  background: #000000;
+}
+
+.header-title a{
+  color: lime;
+}
+
+.head-account-info .name{
+  color: red;
+  font-size: large;
+}
+
+body{
+  background: #000000;
+}
+
+
+.head-account-info.is-logged-in .collapsible li{
+  background: #666;
+}
+
+#settings-list{
+  border:1px solid black;
+  border-top: none;
+}
+
+.card{
+ background: #4b4b4b;
+ color:#7fff69;
+}
+
+.comment-section{
+  background: #051700;
+  color: #00ffe0;
+  border: 1px solid black;
+}
+
+.comment-name{
+  color: #3bff00;
+}
+
+input.comment-submit {
+  color: #046300;
+}
+
+input.comment-input
+{
+  color: red;
+}
+
+a
+{
+  color: green;
+}
+
+div.dark-section
+{
+  background: #000e00
+}
+
+.end-of-feed{
+  background: #074c00 !important;
+  color: #770000;
+}
+
+.comment-message code {
+  background: #404040;
+  color: hsl(139, 53%, 43%);
+}
--- a/dwitter/static/theme_green.css
+++ b/dwitter/static/theme_green.css
@@ -106,7 +106,7 @@ body{
 
 
 .head-account-info.is-logged-in .collapsible li{
-  background: #666;
+  background: #042b00;
 }
 
 #settings-list{
--- a/dwitter/static/theme_green.css
+++ b/dwitter/static/theme_green.css
@@ -78,8 +78,6 @@ ul.messages li{
   border: none;
 }
 
-/* Dark-theme-specific properties */
-
 .head-menu a {
   color: #17ff00;
 }
--- a/dwitter/static/theme_green.css
+++ b/dwitter/static/theme_green.css
@@ -54,7 +54,7 @@ ul.messages li{
 }
 
 .head-account-info.is-logged-in:hover {
-  background: #00ff18;
+  background: #141d14;
 }
 
 .head-account-info .collapsible li {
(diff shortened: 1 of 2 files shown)


dwitter/templates/base.html
```diff
@@ -37,27 +37,36 @@
...
-            localStorage.setItem('theme','');
-          } else {
-            localStorage.setItem('theme','dark');
+          if (theme == '0') {
+            localStorage.setItem('theme','1');
+          } else if(theme == '1') {
+            localStorage.setItem('theme','2');
+          }
+          else {
+            localStorage.setItem('theme','0');
```
joeysino: Keep it simple:
```js
localStorage.setItem('theme', theme);
```

If they set an invalid value, we can deal with it at load time.

dwitter/static/theme_green.css
```diff
@@ -0,0 +1,161 @@
...
+.comment-name{
+  color: #3bff00;
+}
+
+input.comment-submit {
+  color: #046300;
+}
+
+input.comment-input
+{
```
joeysino: This file could use a little cleanup. Perhaps try passing it through [`css-beautify -s 2`](https://www.npmjs.com/package/js-beautify#css--html)

dwitterer:
Also, IMHO the color scheme could be vastly improved. The reds don't go very well with the greens, and the greens you chose look quite... dull :D 

Perhaps colours like #05ad00 or #048900 would look better ¯\_(ツ)_/¯

dwitter/templates/base.html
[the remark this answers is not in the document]
kipkat: Oh! Thank you!

kipkat:
@dwitterer I made the colors nicer. _Also BetterDiscord is against Discord ToS._
image

kipkat:
Only the account name is red. I will fix the awesome button later. u(t) is blue to make a difference between the code inside the function and the function. Also [BetterDiscord is not allowed.](https://www.reddit.com/r/discordapp/comments/6erhdj/so_discord_doesnt_allow_betterdiscord_to_be_used/dicjjos/)

kipkat:
Added Advanced Math Functions too for experts. What do you think? Prefix: $. Their names are very unique and they will not break any dweets.

lionleaf:
Nice progress on the theme.

However, I won't add new math functions or utilities to the "dwitter language". It would drastically change what is possible in a dweet and make the code even further from vanilla javascript. If I were to enable more complex dweets I'd rather increase the character limit. (which isn't happening either). See #95 for deeper discussion. 
You are not the first to attempt to add more functions, I should really add some contributing guidelines so I don't waste the time of more people on this. 

PS: For future reference, if you want to add a two different things, it's better to make them into different pull requests. Add a new pull request, you have to make new branch on your fork, add your new feature there, and you can then open a pull request from that branch. I'd be happy to help on discord.

kipkat:
Okay @lionleaf . Sorry I don't know how the pull requests work. I will reverse the math functions changes.

dwitter/templates/base.html
```diff
@@ -37,23 +37,24 @@
...
+            localStorage.setItem('theme','default');
           }
           location.reload();
           // Call this function in the li to change the mode
         }
 
         // Do the same thing outside the function without changing localStorage to set correct theme
         var theme = localStorage.getItem('theme');
-        if (theme == 'dark'){
+        if (theme != 'default'){
```
joeysino: We should probably make this

```js
if (theme && theme !== 'default') {
```

because I believe `''` is what new users get to start with. (Making `''` and `'default'` the same thing.)

kipkat: That's actually a very good idea!

joeysino: I think you may have edited the wrong bit of code. 🍭

dwitter/templates/base.html
```diff
@@ -37,23 +37,25 @@
...
           // Call this function in the li to change the mode
         }
 
         // Do the same thing outside the function without changing localStorage to set correct theme
         var theme = localStorage.getItem('theme');
-        if (theme == 'dark'){
+        if (theme && theme !== 'default') {
           var l = document.createElement('link');
-          l.setAttribute('href', '{% static "theme_dark.css" %}');
+          l.setAttribute('href', '{% static "theme_'+theme+'.css" %}');
```
lionleaf: Unfortunately this doesn't work.
The {%  %} template functions are run on the server and returns a string that gets included in the javascript. So, it doesn't have access to any javascript variables.

I think you'll have to either go back to redundant if statements, or make a dictionary to get the path values:

themePaths = {''dark": '{% static "theme_dark.css" %}',
                          "green": '{% static "theme_green.css" %}'}
and then 
l.setAttribute('href', themePaths[theme]);

Or something similar.

NOTE: not tested

PS: You should set up dwitter locally so you can test functionality

kipkat: I tried to host it but no success... I was expecting that will not work. I will edit it.

dwitter/templates/base.html
```diff
@@ -83,7 +85,7 @@ <h2 class=header-title>
               <li><a href="{% url 'user_feed' url_username=request.user.username %}"> My profile </a></li>
               <li><a href="{% url 'user_liked' url_username=request.user.username %}"> My awesomed dweets </a></li>
               <li><a href="{% url 'user_settings' url_username=request.user.username %}"> Settings </a></li>
-              <li><a href="#" onclick="switchTheme();return false;">Switch theme</a></li>
+              <li><a href="#" onclick="switchTheme();return false;">Shift themes</a></li>
```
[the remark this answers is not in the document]
kipkat: Shift? Doesn't mean [a,b,c,d] if now is c then d is next, then if now is d then a is next... and so on.

dwitter/templates/base.html
```diff
@@ -83,7 +86,7 @@ <h2 class=header-title>
               <li><a href="{% url 'user_feed' url_username=request.user.username %}"> My profile </a></li>
               <li><a href="{% url 'user_liked' url_username=request.user.username %}"> My awesomed dweets </a></li>
               <li><a href="{% url 'user_settings' url_username=request.user.username %}"> Settings </a></li>
-              <li><a href="#" onclick="switchTheme();return false;">Switch theme</a></li>
+              <li><a href="#" onclick="switchTheme();return false;">Next themes</a></li>
```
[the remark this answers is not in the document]
kipkat: Fixed

dwitter/templates/base.html
```diff
@@ -37,23 +37,26 @@
...
       if (window.localStorage){
         var switchTheme = function(){
-          // Change the imported stylesheet from main.css to main_dark.css or vice versa
-          // Change localstorage theme to 'dark'
+          // Shift themes
           var theme = localStorage.getItem('theme');
-          if (theme == 'dark') {
-            localStorage.setItem('theme','');
-          } else {
+          if (theme !== '') {
```
joeysino: I think this should be:
```js
if (theme === '') {
```
or perhaps simpler:
```js
if (!theme) {
```

dwitter/templates/base.html
```diff
@@ -37,23 +37,26 @@
...
+            localStorage.setItem('theme','');
           }
           location.reload();
           // Call this function in the li to change the mode
         }
 
         // Do the same thing outside the function without changing localStorage to set correct theme
         var theme = localStorage.getItem('theme');
-        if (theme == 'dark'){
+        if (theme !== '') {
```
joeysino: This is fine. But you could also write `if (theme) {`

Idiomatic Javascript is lazy!

kipkat:
I set up Dwitter on my localhost and this modification works perfectly.

dwitter/templates/base.html
```diff
@@ -37,23 +37,26 @@
...
       if (window.localStorage){
         var switchTheme = function(){
-          // Change the imported stylesheet from main.css to main_dark.css or vice versa
-          // Change localstorage theme to 'dark'
+          // Shift themes
           var theme = localStorage.getItem('theme');
-          if (theme == 'dark') {
-            localStorage.setItem('theme','');
-          } else {
+          if (theme === '') {
```
joeytwiddle: Make this:
```js
if (theme) {
```
then it will work as expected when theme is `''` and also when theme is `undefined` (which is what happens the very first time.)

kipkat: ✅

PR #365, opened 2018-08-02
Browserglobals fix

joeytwiddle:
The problem was that sometimes Firefox would call `postMessage()` after the first `<script>` had loaded, but _before_ the second `<script>` had loaded.

If this happened, then `reset()` would break because `browserGlobals` had not yet been built.

This fix starts `browserGlobals` off as null, and only starts using it for cleanup once it has been built (which is now the last thing done when the page loads).

I also re-enabled `browserGlobals` logging in case there are any more problems in future. We can turn it off in a couple of months, if everything is working fine.

Tested:
- Dweets play in Firefox and Chrome
- New Dweet works in Firefox and Chrome

Not tested:
- Scrolling to second page of dweets. (My dev machine only had 5 dweets in the DB right now!)

Note: The first commit was more thorough, but probably overkill for editing. Let's KISS for now, with the second commit.

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -28,7 +28,9 @@
       var dweetId;
       var username;
       var minGIFWidth = 640;
-    
+
+      var browserGlobals = {};
+
       function receiveMessage(event){
         var origin = event.origin || event.originalEvent.origin;
         console.log("Received message from " + origin);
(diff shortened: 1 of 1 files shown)


lionleaf:
This doesn't seem to work locally for me. Where have you tested it?

screen shot 2018-08-03 at 7 43 45 pm

My current errors ^ . 

I suspect we should get some linting going and increase the quality of the dweet code. Seems a bit brittle.

PR #366, opened 2018-08-05
Increase loopbuster timeout to 4 seconds (from half a second)

joeytwiddle:
This should enable users on slow machines/devices to view some of the heavier dweets.

At the same time, it will not leave their device completely unresponsive. After 4 seconds, the dweet gets killed.

I was tempted to go to 8 seconds, but managed to restrain myself. I guess very few dweets need more than 4 seconds.

Having any limit at all is great for users.

### DIFF
--- a/dwitter/static/libs/loopstopper.js
+++ b/dwitter/static/libs/loopstopper.js
@@ -2,7 +2,7 @@ window.stopper = {
 
   loops: [],
   loopTimers: [],
-  TIME_OUT_LENGTH: 500,
+  TIME_OUT_LENGTH: 4000,
   states: Object.freeze({ 
     UNMONITORED: 0,
     MONITORED: 1,


sigvef:
4 seconds feels like a good value to me!

joeytwiddle:
My Asus 453 laptop has two Celeron CPUs at 2.16GHz. I have Chrome running with 15 extensions on Ubuntu.

Chrome gets pretty laggy when there are heavy dweets on the page. But I am happy to accept that. Since at present the alternative is being unable to see some of the dweets at all.

If someone put forward an argument for 2 seconds, I would not object. (But I think my Chrome would still lag anyway!)

(I think the ultimate solution might be an opt-out of loopbusting for busted dweets, or a client-configurable timeout.)

lionleaf:
I think we should find a value and stick to it. It then becomes part of the ruleset as it were "it has to run no slower x fps". I kinda like the idea of 1 second, which gives us 1 fps. Slower than that gets really painful. Though I can be convinced of both 2 sec and 4 sec. But not more.

joeytwiddle:
To clarify: I'm only suggesting an opt-out for **viewers**, not for publishers.

I think the philosophy of a fixed ruleset makes sense. It's just that pesky framerate which varies between machines.

Currently, if 5 people on PCs make dweets just within the threshold, then 5 viewers on mobile will see only timeout messages!

Personally I am very curious to see what a slow dweet renders :rainbow:, but then I scroll on by. <-- That is the situation I'm hoping to reach.

Since an opt-out would mean more UI and more code, perhaps more simply, we could enforce a 1 second threshold for editing/publishing, but allow a 4 second threshold when viewing.

lionleaf:
An opt-out on a per dweet basis might be a pretty good solution if implemented correctly. "Your device was not able to show this dweet. If you really want you can *_try harder_*." pop-up: "running this dweet without a timeout. This can hang or crash your browser. Continue?"

lionleaf:
This discussion should really move to an issue rather than in a closed pull request

PR #377, opened 2018-08-11
Improvements to kipkat's green theme

joeytwiddle:
Best shown in pictures (before, after):

 

I made a separate branch, because I reformatted some JS and CSS, and it's impolite to push heavy changes onto someone else's branch!

Note that if this PR gets merged, then GitHub will automatically merge the original PR too, because it is a subset.

kipkat if you like these changes, but want to keep working on them, you can pull from this, or easier, I can push onto your branch.

----

Idealistic thoughts for the future:

- It might be nice to do some cleanup of the HTML before we add more themes. We have a few different types of "buttons" and a couple of types of "links" which could probably be unified and made more semantic.
- Some CSS rules are duplicated, and could be made more general.
- We could also use CSS variables, so a few colour changes can propagate over the whole site more easily.

### DIFF
--- a/dwitter/static/theme_dark.css
+++ b/dwitter/static/theme_dark.css
@@ -2,25 +2,29 @@ textarea {
   border: none;
 }
 
-ul.messages li{
-  border:1px solid black;
+ul.messages li {
+  border: 1px solid black;
 }
 
 .content-div {
-  background:#4b4b4b;
+  background: #4b4b4b;
   color: white;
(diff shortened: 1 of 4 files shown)


kipkat:
Wow. Thx, it's nicer.

joeytwiddle:
Since kitkat approved of these changes, I will push them directly onto his branch, and close this PR.

PR #378, opened 2018-08-15
Upgrade to Django 1.11

stianjensen:
Django 1.11 is the next LTS release of Django and the last release to
support Python 2. Luckily, Django 1.11 will be supported with security
updates until sometime in 2020 so there's no immediate rush to rewrite
everything to Python 3 just yet.

Still, Django 1.8 will soon stop getting security updates, so upgrading
to Django 1.11 now means it will still be easy to take in new Django
Security fixes.

Migrating was fairly straight-forward, and involved reading the release
notes of Django 1.9, 1.10 and 1.11, as well as upgrading any third-party
libs that had released compatibility updates for Django 1.11:

- https://docs.djangoproject.com/en/1.11/releases/1.9/
- https://docs.djangoproject.com/en/1.11/releases/1.10/
- https://docs.djangoproject.com/en/1.11/releases/1.11/

The most tricky part to fix was django-subdomains which has been unmaintained for a long time. Luckily, I found a well-maintained replacement. This part of the upgrade should get extra attention so nothing breaks. I'm unsure which parts of the codebase are untouched by the automatic tests, so I guess all parts of the site should be manually tested, for safety before deploying.


stianjensen:
> Hang on, I definitely forgot to restore some now-relative urls to absolute.

Fixed.

stianjensen:
We probably want to set `HOST_SCHEME` to 'https' in local.py in production. The default is '//' which gives urls that look like this: `//dweet.dwitter.com/id/1`. It's convenient locally, since it means urls will work with and without https, but they do look a bit strange for permalinks.

Also, I had to set ALLOWED_HOSTS to specific values now. I think it's a restriction in one of the django-versions after 1.8. We probably need to set it to something in production as well, although that may be done already (ie. it's only with DEBUG=True that it wasn't mandatory before).

PR #379, opened 2018-08-16
Improved green theme + other minor theme patches

dwitterer:
See the descriptions of my commits for more details, but basically, I have:
* Restored some gray areas to the dark theme that seem to have been lost (or I don't know, it's just a bit nicer now)
* Made the green theme more visually appealing
* Passed all the CSS through not one, but *two* beautifiers, so now it's sexy af
* Fixed a little thing i saw in the comments of dweet.html (i.e that they were split)

But anyway...

I propose a green theme like this:
<pre align=center></pre>
Screenshot 2
Screenshot 3
Screenshot 4
Screenshot 5

I'll let @katkip and @lionleaf decide, but I think this looks much more user-friendly and consistent while maintaining a nice "hackery" theme. As for the theme switcher, I plan on making that a submenu or (for a more mobile-friendly solution) an external page, so we don't need to worry about that too much for now...

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -1,737 +1,776 @@
 * {
-  border:0px;
-  margin:0px;
-  padding:0px;
+    border: 0px;
+    margin: 0px;
+    padding: 0px;
 }
 
 .clear-fix:after {
-  content: "";
-  display: table;
-  clear: both;
+    content: "";
+    display: table;
+    clear: both;
 }
 
 .avatar {
-  width: 36px;
-  height: 36px;
-  border-radius: 18px;
-  background: #ccc no-repeat center;
-  background-size: cover;
-  overflow: hidden;
-  display: inline-block;
-  vertical-align: middle;
-  box-shadow: 0 0 2px rgba(0, 0, 0, 0.2);
+    width: 36px;
+    height: 36px;
+    border-radius: 18px;
+    background: #ccc no-repeat center;
+    background-size: cover;
+    overflow: hidden;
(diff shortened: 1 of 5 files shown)


dwitterer:
And since I am here, I'd like to propose another idea I had about rounded corners on `c`, which I believe would make the whiteness of the canvasses integrate much better with the dark themes. It looks like this: 
[Screenshot](https://i.imgur.com/gNGCPnJ.png). So I'm interested to hear your input on this...

lionleaf:
The improvements look good to me, but you've got some conflicts.

Similar to stianj I prefer to stick to square corners

dwitterer:
screenshot from 2018-08-22 00 41 55

Changed the background to be black, as per kipkat's request... What do you think? Also made the borders lime for visibility, but they might be a bit *too* visible now - I'll look into this tommorrow if it's an issue /shrug

Fun fact: It won't show up here, but for some reason my linux has a different color profile, and it makes the limes less jarring, and the theme looks much better like that :thinking:

lionleaf:
Will merge if it runs well and @kipkat approves.

kipkat:
The colors you used are ugly. The theme is 102% different. You can create a new ugreen_theme.css if you want.

lionleaf:
Even though there isn't really an ownership of these themes after they've been merged, in this case I say we keep it as is, as it appears these changes goes against the theme kipkat wanted to contribute. You're more than welcome to keep working together on tweaking it of course, but closing this for now.

PR #381, opened 2018-08-21
Change first comment placeholder text

lionleaf:
screen shot 2018-08-20 at 8 27 01 pm

As suggested by @joeytwiddle I'd like to change the placeholder in the new "first comment" text field from `Comment` to something else. Rather than having the discussion in an issue, I made a pull request and we can discuss here.

My current suggestion is `Add a caption... (optional)`, but I'm open to changing it.

Some other alternatives:
screen shot 2018-08-20 at 8 32 54 pm
screen shot 2018-08-20 at 8 32 22 pm
screen shot 2018-08-20 at 8 31 59 pm
screen shot 2018-08-20 at 8 31 36 pm
screen shot 2018-08-20 at 8 31 08 pm

I think the word choice here will have a decent impact on what kind of comments we will get.
I like `caption` because it suggests both a title and/or a description.

Thoughts?

### DIFF
--- a/dwitter/templates/snippets/dweet_card.html
+++ b/dwitter/templates/snippets/dweet_card.html
@@ -101,7 +101,7 @@
             {% if not embed %}
             {% if request.user.is_authenticated %}
             <div class=post-dweet-div>
-                <input type=text placeholder="Comment" name="first-comment" class="new-dweet-comment-input" />
+                <input type=text placeholder="Add a caption..." name="first-comment" class="new-dweet-comment-input" />
                 <input
                     class="remix-button"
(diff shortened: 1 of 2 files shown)


joeytwiddle:
My vote goes for "Add a caption..."

I think you could leave out "(optional)" since that is somewhat implied by the button not being disabled.

My second vote goes for "Descriptive title for your dweet..." 😉

lionleaf:
I agree; I like "Add a caption..." 
screen shot 2018-08-21 at 6 07 26 pm

Approve if you agree and I'll merge it

PR #382, opened 2018-08-22
Run linter on main.css

lionleaf:
Split this into a two pull requests as there's still some work on the green theme. This should be safe though


kipkat:
image
image
image
image


0 errors, 49 warnings.

PR #383, opened 2018-08-22
Lint theme_green.css

lionleaf:
Split this out in case @dwitterer would prefer to wait with merging this until #379 is merged (and I would update it after that of course)


kipkat:
image
image

Only 19 warnings for me.

PR #385, opened 2018-08-24
Aos

kipkat:
Simple zoom-in Animation On Scroll. It doesn't affect performance (even if the canvas has 1 fps) and looks dope.

- **It was not tested on a local host, only on the browser dev tools modification.**
- **It is [external](https://github.com/michalsnik/aos)! I'm not sure where if we must provide credits.**

### DIFF
--- a/dwitter/templates/base.html
+++ b/dwitter/templates/base.html
@@ -5,6 +5,7 @@
     <meta name="description" content="{% block meta_description %}{% endblock %}">
     <title>{% block title %} Dwitter {% endblock %}</title>
     <link rel="shortcut icon" href="static/favicon.ico" type="image/x-icon">
+    <link href="https://unpkg.com/aos@2.3.1/dist/aos.css" rel="stylesheet">
     <link rel="icon" href="static/favicon.ico" type="image/x-icon">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     {% load compress %}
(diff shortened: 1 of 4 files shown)


dwitterer:
Ooh, pretty! :heart_eyes_cat:

lionleaf:
Sorry about the effort you put into this, but I'm not a fan of animations like that.

kipkat:
It makes the interface more user-friendly.

dwitterer:
It's a bit easier on the eye... TBH a long feed of data can be overwhelming at times. Plus it makes the website a bit more... alive - something that dwitter could really use :/

So I think I'd disagree with you there @lionleaf...

kipkat:
Yeah, probably @lionleaf isn't a fan of this kind of animations, but the community will like them. He is the creator so he is deciding. Poll [here](https://www.strawpoll.me/16341585).

lionleaf:
I prefer to keep the UI of dwitter as clean as possible, and pop-in animations on dweets is, unfortunately, not something I want. I appreciate that you put effort into this, but in this case I won't be merging this. To avoid similar things in the future I suggest you check with me directly, or open an issue, if you're unsure whether the feature/change is something I want. There are plenty of things I'd love some contributions for, and it's a shame to have dwitter work go to waste. 

Definitely check with me if it changes the main UI/UX.

PR #386, opened 2018-08-26
Python3

stianjensen:

I did this, but it caused changes in files I haven't touched, so I assume they are currently not linted properly on master

No open issues for python3 support that I could find.

---

I haven't extensively tried to test or check that all third-party packages work with this, but new relic was the only one that caused errors upon boot.
One remaining thing that may be done is to run the server locally with deprecation warnings enabled, to see if something should be changed.

We might want to make some more changes to the readme to explain that python3 is required instead of python2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: python
 python:
       - "2.7"
       - "3.6"
-matrix:
-      allow_failures:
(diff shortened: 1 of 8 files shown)


stianjensen:
Also, I guess this should update the travis config.

Issue #375, opened 2018-08-08
No CORS headers for API requests


joeytwiddle:
If we want to be able to access the API from other websites, then we'll need something like this to work:

```js
fetch("//www.dwitter.net/api/dweets/?limit=1", {mode: 'cors'}).then(res => res.json()).then(console.log)
```

That does work fine if I run it in the browser console on dwitter.net. But from any other site, I get the error:
```
No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

This issue was raised in Discord by kipkat.

PR #387, opened 2018-08-27
Enables CORS

BalintCsala:


kipkat:
The thing was to enable it on /api/, not on the whole website. Now everybody can cheat using external sources.

stianjensen:
You need to add `django-cors-headers==2.4.0` to requirements.txt, in order to make this work.

stianjensen:
> The thing was to enable it on /api/, not on the whole website. Now everybody can cheat using external sources.

It's probably fine to limit CORS to only the API, but I don't see how enabling CORS generally opens for cheating in dweets?

kipkat:
@stianjensen see [h/external](https://www.dwitter.net/h/external) comments and scripts. that's you you cheat and that's why CORS is enabled.

stianjensen:
> see h/external comments and scripts. that's you you cheat and that's why CORS is enabled.

Right. That's blocked by CSP, Content-Security-Policy, which works by having response headers on the initial page load say which domains are okay to load resources embedded on dweet.dwitter.net from, i.e. saying no dweets can load content from the codepen.io domain.

CORS means Cross-Origin Resource Sharing, and is also response headers sent by the server, but in this case it says which other pages are allowed to load resources _from_ the dwitter.net domain, i.e. blocking or allowing mysite.com to load all the comments for a specific dweet-id.

kipkat:
Issue that will get fixed: #375

kipkat:
Still doesn't work.
```Failed to load https://www.dwitter.net/api/dweets/?limit=1: Request header field Access-Control-Allow-Origin is not allowed by Access-Control-Allow-Headers in preflight response.```
Live example: https://kipkat.github.io/Dweet-Finder/

**Comment by joeytwiddle** [comment]
It works for me, from the browser console on kipkat's site:

```js
> fetch("//www.dwitter.net/api/dweets/?limit=1", {mode: 'cors'}).then(res => res.json()).then(console.log)
{count: 8266, next: "https://www.dwitter.net/api/dweets/?limit=1&offset=1", ...}
```

I don't know why kipkat's XHR isn't working. As you can see from the Network tab, it actually sends a `OPTIONS` preflight, which seems to work, but it never sends the main request.

Please just use `fetch()`, like everyone else does in 2018!

stianjensen:
The concrete failure you're getting now is from this line, which I think you can just remove: `request.setRequestHeader("Access-Control-Allow-Origin", '*');

But as @joeytwiddle says, it's probably easier to just use the fetch API.

Issue #330, opened 2018-04-22
Gif record button doesn't have a similar hover effect as the other links


lionleaf:
Makes it inconsistent with the rest of the UI:

<img width="129" alt="screen shot 2018-04-22 at 12 28 13 pm" src="https://user-images.githubusercontent.com/610925/39094524-aa43c968-4628-11e8-8c43-b73230e2e65c.png">

lionleaf:
Same with the `delete` button.

kipkat:
```css
.record-text:hover, input.delete-button:hover{
  text-decoration:underline;
}
```

lionleaf:
Closed by #390

PR #390, opened 2018-09-04
Hover effects

leventdeniz:
Fixes issue #330 on hover effects.

(this is my first pull request, I hope I did everything right)

### DIFF
--- a/dwitter/static/main.css
+++ b/dwitter/static/main.css
@@ -108,7 +108,6 @@ a.button:visited {
   color: inherit;
 }
 
-
 body {
   background: #f5f5f5;
   font-family: 'Roboto', sans-serif;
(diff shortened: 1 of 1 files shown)


kipkat:
That might be a good idea. I asked @lionleaf if we can use [Hover.css](http://ianlunn.github.io/Hover/) in the navigation bar and he didn't answer.

leventdeniz:
[Hover.css](http://ianlunn.github.io/Hover/) may be too much, since we have these minimalistic underline hover-effects and @lionleaf might want to keep this style.
But the "new dweet"-button should get a proper hover-effect.

kipkat:
We can use the css keystokes to make them look modern.

lionleaf:
Thanks! Works like a charm

leventdeniz:
Awesome!

PR #391, opened 2018-09-06
Textexp

kipkat:
Simple text expander to make a dweet more easy.
- [x] Lint

### DIFF
--- a/dwitter/static/js/expander.js
+++ b/dwitter/static/js/expander.js
@@ -1,0 +1,34 @@
+editor = document.getElementById("editor")
+function u_repl(str,old_,new_){return str.split(old_).join(new_)}
+function expand(c)
+{
+    var start = c.selectionStart,
+    	end = c.selectionEnd;
+    c.value = u_repl(c.value,"hsl;","`hsl(${},99%,50%)`;")
+    c.value = u_repl(c.value,"hsla;","`hsla(${},99%,50%,.5)`;")
+    c.value = u_repl(c.value,"x.fs=","x.fillStyle=")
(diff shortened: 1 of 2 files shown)


lionleaf:
As mentioned on discord, I think having fully automatic autocomplete might surprise people. If we're adding this feature I think it should be triggered by a keystroke, tab being the most obvious one. Can you make it into a tab-completion feature? 

You might have to change the behaviour slightly for that. For instance,

x.f => x.fill,  x.fillR => x.fillRect etc.

This might be slightly misleading as people might expect a full tab completion for all the members (rather than hardcoded), but is still quite useful.

That being said, dweets are very short already, so typing speed isn't really a major issue.

What do other frequent dweeters think of this?

kipkat:
I set the expand key to ` ` (1 space) because it is the best choice here. It is easy and I'm sure ; + space isn't so frequent. It can be changed in the localStorage (key: "_expandKey_"). The ``x.fr = x.fillr = x.fillRe = x.fillRect`` is pretty tricky here... I think that the autocomplete feature isn't so good idea here because we need _strict_ snippets to avoid misexpands (**example: x.f = x.fillRect() or x.fill(), Math.a = Math.abs() or Math.acos()**)

kipkat:
The `tab` trigger can't be used here without massive textbox event modifier script. Tab will only shift the html element selection, for example while editing and pressing tab, it will select the caption textbox. I can make the trigger to `  ` double spacebar. Also there is a bug:

```
    var start = c.selectionStart,
    	end = c.selectionEnd;
    c.setSelectionRange(start, end);
```

This will cancel any selection every call of the function in the loop. The script is also newbie maked, just a prototype and concept. I am going to make a good one the next week.

PR #409, opened 2018-10-19
Fix #361

### LINKED ISSUE #361: /random - all awesomeness is resetting to 1
### In /random, the score indicator _(class="score-text")_ becomes 1 when visiting the /random page (until you like the dweet.)
**More information:**
- Even if you liked it before visiting /random, it still resets
- If you un-like it after that, it does not reset
Probably it is a database error or something, but I created an easy fix (temporary):
```js
b = document.getElementsByClassName("like-button")
for(let x=0; x<b.length; x++){b[x].click();b[x].click()}
```
Paste that in the Developer Web Console to make fix the visible/loaded ones.

prplz:
Fixes 'like' counting on the random page

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -59,7 +59,8 @@ class DweetFeed(ListView):
             .select_related('author')
             .select_related('reply_to')
             .select_related('reply_to__author')
-            .prefetch_related('comments'))
+            .prefetch_related('comments')
+            .prefetch_related('likes'))
 
         return queryset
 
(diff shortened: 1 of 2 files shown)


lionleaf:
I still like the idea that came up in discussing this issue of having a deterministic random order based on a url parameter. That way your browser history can get you back to a specific random sort that would be very hard to find otherwise. Not sure how I would achieve that, though. Can you seed the '?' sort?

stianjensen:
Navigating to /random can give you a http redirect to /random?seed=specificseed for instance.

lionleaf:
Yes; but how would you feed the seed to the random sorting?

PR #415, opened 2018-10-21
Cap animation to 60hz

### LINKED ISSUE #414: Bug: Playback speed tied to monitor refresh rate
I have a 165hz monitor and dweets run at nearly three times the normal speed on it. I have to check them on my phone to see what they are supposed to look like.

pavelb:
Fixes #414


lionleaf:
Thanks for the fix and the analysis and testing in #414 . I will trust your judgment that average 60fps case with slightly uneven frame timing works best 👍 

Merging and deploying.

PR #416, opened 2018-10-21
Add an epsilon to the frame timing check

lionleaf:
I was getting frame timeouts when the time was off by 0.3ms etc.
This change also uses the timing passed in by requestAnimationFrame
rather than Date.now. This should be more stable as this is a time
that is sampled just before the browser starts doing all the redraws,
so it doesn't depend on how much compute has been done before the
dweet callback is called.

I think.

I am getting some frame skipping on my macbook even with this, but from the timing of them I think they must be requestAnimationFrame triggering extra frames when scrolling or something.

#414 #415

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -232,7 +232,8 @@
 
       if (playing){
         requestAnimationFrame(loop);
-        if (frame_time < Math.floor(nextFrameMs)) {
+        epsilon = 1.5; // Acounts for different timestamp resolution and slight jitter
+        if (frame_time < nextFrameMs - epsilon) {
           return;  // Skip this cycle as we are animating too quickly.
         }
         nextFrameMs = Math.max(nextFrameMs + 1000 / FPS, frame_time);
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -227,16 +227,15 @@
     var src = (u + "").replace(/(function u\(t\) {|}$)/g, "");
     newCode(src);
 
-    function loop() {
+    function loop(frame_time) {
       stats = stats || createStats();
 
       if (playing){
         requestAnimationFrame(loop);
-        var now = new Date().getTime();
-        if (now < nextFrameMs) {
+        if (frame_time < Math.floor(nextFrameMs)) {
           return;  // Skip this cycle as we are animating too quickly.
         }
-        nextFrameMs = Math.max(nextFrameMs + 1000 / FPS, now);
+        nextFrameMs = Math.max(nextFrameMs + 1000 / FPS, frame_time);
       }
       time = frame/FPS;
       if(time * FPS | 0 == frame - 1){
@@ -289,7 +288,7 @@
       }
     }
     if(autoplay) {
-        loop();
+        loop(0.0);
     }
 
     function reset(){


pavelb:
Adding the epsilon is a bit strange. If you think about it, its the same as

`nextFrameMs = Math.max(nextFrameMs + 1000 / FPS + epsilon, frame_time);`

which means you are effectively targeting a frame rate of `FPS/(1 + epsilon*FPS)` which is about 66 at current numbers. Dweets would run 10% faster on all >60hz monitors. We should use a much smaller number.

Frame skipping doesn't seem possible with the current setup; I must be missing something. We ensure we let through at most one frame every 1/FPS since the start (or reset). Skipping a frame would mean we triggered sooner than 1/FPS since the last one, but the last one was triggered by a successful frame so that means requestAnimationFrame triggered twice in 1/FPS doesn't it? Perhaps this has to do with error accumulation in nextFrameMs? What if we do something like this

`
var frameDurationMs = 1000/FPS;
nextFrameMs = (Math.floor(frame_time / frameDurationMs)+1)*frameDurationMs;
`

I can't test this right now.

How are you profiling frame skipping? The performance tool in Chrome is nice.

joeytwiddle:
I don't think the epsilon changes the resulting FPS. It does allow a frame to render early, but it doesn't change the allowed window for future frames. `nextFrameMs` is still incremented by exactly `1000/FPS`, with no regard for the current time (unless we are over a frame behind).

```js
nextFrameMs = Math.max(nextFrameMs + 1000 / FPS, frame_time);
```

According to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) the `frame_time` is not entirely accurate: some browsers randomize it for security reasons.

Would it be possible to set `epsilon` to 8, or `1000 / FPS / 2`? That would allow early frames for machines with quite wide variation, so they won't lose framerate just because they are firing at irregular times. (It would also help to test the increased/constant FPS theory.)

pavelb:
Ah. Epsilon acts modular modulo 1/FPS because it will let the nextFrameMs run ahead until epsilon is less than 1/FPS. That is why it doesn't affect the long-term target FPS even if you set it to 1 hour. Switching to my numerically stable option would not allow nextFrameMs to run head and would affect the target FPS.

In terms of value, things get a bit confusing. Shouldn't we use the max value? Given the modulo nature the max value is 1/FPS - ε but that's the same as -ε (except for the change in what nextFrameMs represents). That would make the code look like

```
if (frame_time < nextFrameMs + ε) { return; }
```

But isn't that the same as

```
if (frame_time <= nextFrameMs) { return; }
````

That seems identical to the current code. Did my reasoning go wrong somewhere?

lionleaf:
Hmm... I need to think about this one. In any case we need to do some rounding, as nextFrameMs has fractional ms whereas frame_time has a resolution of 1ms. So you might get the comparison of 17 < 16.6667 which fails. So maybe the right solution here is `ceil(nextFrameMs)` instead?

PS: I'm not sure I follow your modulo thinking. What exactly is modulo what? Seems to me it would imply something that runs up to a certain value and then wraps back to 0.

lionleaf:
Changed it to Math.floor. Let me know what you think.

lionleaf:
Thanks, I get what you're saying now.

Unless we're thinking about 500-1000Hz monitors we should be good with an epsilon <2ms or with `Math.floor`. While you're correct in that `Math.floor` is effectively a window that changes every frame in the [0,1)ms range, it makes more sense conceptually due to the potentially different resolution of the timestamps.

In the cases where the given timestamp has a 1ms resolution, that introduces a similar offset in the range [0,1) of significance since 1000/60 is not an integer. `Math.floor` would just take that into account.

Anyway, I think we all agree now that we need something? I'm happy with either `Math.floor` or a small `epsilon`. Given the argument above, I will go for `floor` unless anyone wants to chuck in a vote for `epsilon`

Voting: Hooray for `floor`, thumbs-up for `epsilon`, comment to add more serious discussion :)

lionleaf:
As I switched back to an epsilon, I merged it. Let me know if anyone still have any comments on the code

Issue #334, opened 2018-04-22
Bug in link rendering


iver56:
https://www.dwitter.net/d/435

That rendered comment looks buggy. I think the original comment just had a link there

![bilde](https://user-images.githubusercontent.com/1470603/39097575-73fa392e-465e-11e8-8a57-e8fab54a78c4.png)

lionleaf:
Oh that's kinda nasty. Guess we have to make sure that any hashtag is not part of a valid URL... Hmm.

stianjensen:
Another: https://www.dwitter.net/d/477

joeytwiddle:
How about requiring a word boundary before the `#`? That is `\<#` in some regexp syntaxes, `\b#` in others.

Doing that would ignore the following:
```
abc#def
foo.com/bar#baz
```

But it would still match the following:
```
foo.com/bar.#baz
```

Or if you also wanted to ignore the latter, you could get more strict: require a blank space before the hashtag (or start of string). That would be `(^|\s)#` in some regexp syntaxes.

That would not even match this:
```
file this under:#banana
```

stianjensen:
Do other platforms allow hashtags that are not preceded by a space?

lionleaf:
Looks like it's fairly common to require a space or a special character before a hashtag, but varies per platform. Autocomplete and highlights as you write would solve this no matter which way we go. I guess a preceding space would make life easier, but slightly frustrating for users in some cases!#like_this

lionleaf:
Another arguably false positive for hashtags: https://www.dwitter.net/d/1486

Thanks to fujimotos from discord

atesgoral:
Yeah, I just ran into another scenario where a hex color code inside a code snippet was hyperlinked as a tag: `"#362":"#bb3"`

https://www.dwitter.net/d/9083

kipkat:
I suggest something special for colors: [[hexcolor]].
Javascript version: (I don't know the django version)
```js
comment = `Yeah, I just ran into another scenario where a hex color code inside a code snippet was hyperlinked as a tag: "[[362]]":"[[bb3]]"`
// ALL THE OTHER FILTERS HERE
var newcomment = []
spl=comment.split("[[")
for(i=0;i<spl.length;i++)
{
    if(spl[i].indexOf("]]")>-1)
    {
	spl[i] = "#"+spl[i].replace("]]","")
    }
    // spl[i] is now the color. process it here
    newcomment.push(spl[i])
}

[...]

PR #421, opened 2018-10-26
Hashtags improvement and code block sanitization

dwitterer:
I've worked a bit on the regex behind the hashtags, and the way that code blocks are handled in `dweet_card`s. I changed the regex for hashtags, such that hashtags must now be preceded by either a space, or nothing at all, and must two characters or greater (i.e length >= 2). Hashtags must still begin with a letter, and can only contain letters, numbers, and underscores. However, I have also cleaned up the regex, removing a couple redundant rules\*, so stuff such as the length limitation, or the letter-start rule can be more easily changed.

The reasoning behind the length rule was to prevent accidental hashtags such as in d/2667. I doubt that one-letter hashtags such as "#g" would have been very useful or informative anyway...

\* - Redundant rule was that we had a character range: [a-zA-Z_] followed by another range [a-zA-Z0-9_] and the \* quantifier. I cleaned this up by using the [a-zA-Z0-9_] range with a negative lookbehind for digits.

Now, to change the length limit, all we would need to do is change the `{2,}` to a `*`, and to enable numerical-starting hashtags, we would just need to remove the negative lookbehind `(?<!\d)`...

I have also made it so that all links get removed from code blocks, to keep those clean, and to allow them to be used for escaping hashtags or mentions, like on discord. This, however, now means that we need to do `insert_code_blocks` AFTER `insert_magic_links`, but that's a non-issue...

Fixes (some of?) #334


stianjensen:
> Ok, so now the tests are failing because of the single-character hashtag test. Is there a reason we need to test for that? We don't need single-character hashtags...

Sounds like we don't need that single-char test! But the other failing tests probably need fixing.

dwitterer:
What other failing tests? This is the only one that fails now. I've fixed all the other ones.

dwitterer:
After a discussion on Discord with @lionleaf, I am removing the 2-char minimum limit for hashtags. The problems that one-character hashtags have caused should be fixed anyway, and noone uses single-character hashtags anyway. So, as lionleaf seems to insist on having them, I am adding them back.

Lionleaf now needs to write some tests for the new fixes, merge this in, and then I will open a new PR for numerical hashtags, and hashtags will be good for now... :D

dwitterer:
I've also enabled hashtags beginning with numbers, since the space precedence check makes them safe to use now. Also added an if statement to make sure that those tags need to have at least one letter - this prevents senseless tags like #78992245, and (mainly why I did this) makes it more like twitter's hashtag handling. This extra check can be removed if you don't like it, but otherwise you just need to add the new tests, test my changes out for yourself, and this PR is ready to merge :D!

dwitterer:
Covered, and expanded upon in #424.

PR #427, opened 2018-11-26
Add feeds for top week/month/year/all-time

lionleaf:
Add 4 variants of the top feed; top of the week (last 7 days),
top of the month (last 30 days), top of the year (last 365 days)
and top of all time (the old top behaviour).

Also adds a dropdown menu to select between the new feeds.

And adds new unit tests.
screen shot 2018-11-25 at 11 16 33 pm
screen shot 2018-11-25 at 11 08 23 pm
screen shot 2018-11-25 at 11 08 29 pm
screen shot 2018-11-25 at 11 08 43 pm

Default top feed is "top of the month", which will hopefully be more dynamic than top of all time.

Note; I made a conscious decision to pick a handful of specific feeds (`top/week`, `top/month`, `top/year`, `top/all`) rather than a more flexible url-parameter (eg. `top?days=23`). This gives us fewer "official" rankings with nice URLs that the whole community get the same small handful of lists. I know it's easy to want full configurability; but I'm opting to keep things simple for now :D 

Would be great to get some feedback on the code :)

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -18,18 +18,13 @@ import json
 
 
 class SortMethod:
-    NEW = {'ordering': ['-posted'],
-           'name': 'new'}
-    HOT = {'ordering': ['-hotness', '-posted'],
-           'name': 'hot'}
-    TOP = {'ordering': ['-num_likes', '-posted'],
-           'name': 'top'}
-    RANDOM = {'ordering': ['?'],
-              'name': 'random'}
+    NEW = {'ordering': ['-posted']}
+    HOT = {'ordering': ['-hotness', '-posted']}
+    TOP = {'ordering': ['-num_likes', '-posted']}
+    RANDOM = {'ordering': ['?']}
 
 
 class DweetFeed(ListView):
-
     sort = SortMethod.NEW  # default sort by new
     model = Dweet
     context_object_name = 'dweets'
@@ -46,11 +41,12 @@ class DweetFeed(ListView):
         context['title'] = self.title
         context['feed_type'] = self.feed_type
         context['show_submit_box'] = self.show_submit_box
-        context['sort'] = self.sort['name']
+        context['feed_name'] = self.feed_name
         return context
 
     def get_queryset(self):
         queryset = self.get_dweet_list()
+        queryset = queryset.annotate(num_likes=Count('likes'))
         queryset = queryset.order_by(*self.get_ordering())
 
         # Optimize the SQL query:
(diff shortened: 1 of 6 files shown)


dwitter/feed/views.py
```diff
@@ -169,20 +230,23 @@ def get_dweet_list(self):
...
     sort = SortMethod.NEW
+    feed_name = "new"
 
     def get_title(self):
         return "Dweets by u/" + self.kwargs['url_username'] + " | Dwitter"
 
 
 class TopUserFeed(UserFeed):
     sort = SortMethod.TOP
+    feed_name = "top"
```
stianjensen: How will this view handle not having options for week, month etc.?

lionleaf: Just fine; the whole top menu has a separate code path for the user menu, the hashtag menu and the dweet feed menu; since the links need to be different.

I did catch a css bug thanks to your comment though, the "top" feed of a user didn't correctly bold "top". Fixed.

lionleaf:
> View code has some possibilities for more sharing of code between "top" view classes, but it's not important to fix for now, imo.

Yeah, I know I could reduce the code reuse significantly there; but I didn't want to go overboard with sub-classes; opting for more repitions in the hope that it might be more straight-forward to understand for other people dabbling in dwitter. That said; I'd love to get your take on the best balance and how you would do that? I was considering a new "TopDweetFeed" class with the common new parts (context etc) and then subclass that.

stianjensen:
> Yeah, I know I could reduce the code reuse significantly there; but I didn't want to go overboard with sub-classes; opting for more repitions in the hope that it might be more straight-forward to understand for other people dabbling in dwitter. That said; I'd love to get your take on the best balance and how you would do that? I was considering a new "TopDweetFeed" class with the common new parts (context etc) and then subclass that.

One thing I was looking at was that the num_likes annotation was repeated across both top and non-top feeds, and is basically needed in all views. So somehow reworking the get_dweet_list helper on the base class to always annotate, and make the other views just attach more filtering to that instead of starting from `Dweet.objects` could be nice.

lionleaf:
Did a little refactoring and cleanup of feed/views.py . What do you think @stianjensen ?

dwitter/static/main.css
```diff
@@ -461,6 +461,30 @@ iframe {
...
+  opacity: 0;
+  height: 0;
+}
+
+.top-sort-list li:hover>.collapsible,
+.top-sort-list li .collapsible:hover {
+  position: absolute;
+  display: block;
+  opacity: 1;
+  transition: opacity .1s ease-in-out, margin .1s step-start;
```
stianjensen: I think you need to add a `margin-top: 0` here, and a `margin-top: -9999px;` in the block above.
Currently, the dropdown shows up when the mouse cursor passes below the top navigation bar.

lionleaf: Thanks! You're right, that would've been really annoying. Nice catch.

Your suggestion fixes it, see new commit.

stianjensen:
LGTM, ship it!

dwitter/static/main.css
```diff
@@ -475,6 +475,7 @@ iframe {
   transition: opacity .1s ease-in-out, margin .1s step-end;
   opacity: 0;
   height: 0;
+  margin-top: -9999px;  # remove the hover-trigger while invisible
```
stianjensen: Actually, this isn't valid css comment syntax, but right here it seems like browsers just ignore it anyway, so it doesn't break anything.

`/* remove the hover-trugger while invisible */` would be correct.

lionleaf: Fixed

PR #434, opened 2019-01-10
Fix white-on-white in dark themed top-sort-list

sethjust:
In Firefox 64.0 with the dark theme, the menu for the top dweet's period is rendered in white-on-white. A single CSS rule sets the background appropriately (to match the header).

### DIFF
--- a/dwitter/static/theme_dark.css
+++ b/dwitter/static/theme_dark.css
@@ -61,6 +61,10 @@ ul.messages li {
   color: white;
 }
 
+.top-sort-list a {
+  background: #4b4b4b;
+}
+
 .like-button {
   border: 1px solid black;
 }


dwitterer:
There shouldn't be any white backgrounds or borders in the dark theme anyway... It seems that when a PR was merged, which aimed to simplify the CSS for that theme by removing duplicate and redundant rules,  some of my original rules were lost. It would be great if someone could go through and see if anything's out of place.

sethjust:
I've been using the dark theme almost exclusively and haven't encountered any other bad backgrounds. I think this was just an oversight in the implementation of the new top-period selection.

dwitterer:
Found some:
* "Loading..." card
* Links in user menu (when you click on the profile image) should probably be white too
* I don't know if we should remove the logo from the dark theme, but it looks kind of weird with the white background - maybe round the corners?

If you could fix those too, that would be great...

dwitterer:
Also, green theme has red username on the top bar (likely unintended), and the text in the user menu is gray, making it kind of hard too see. We should make those lime green, like the rest of the page.

PR #437, opened 2019-02-19
Add simple report flag to dweets and comments

sigvef:
This adds simple report flags to dweets and comments. Reports are sent
to moderators on Discord. Injecting more complicated HTML through
js/ajax for the "load more comments" section turned out to be a little
complex w.r.t. CRSF. Then I realized that we rarely have enough comments
to trigger a "Load more comments" link, and when it is there, it often
hides just a single comment. Therefore, I figured we can just remove the
"Load more comments" entirely.

In order to smoothly place the report button, I did a small visual
refresh of the comments as well.

---

- [X] Run `make lint` to ensure that all the files are formatted and using best practices.
- [X] Link to any issues this PR is solving.


dwitter/templates/base.html
```diff
@@ -29,6 +29,8 @@
     <link rel=stylesheet href='{% static "main.css" %}' >
     {% endcompress %}
 
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.7.2/css/all.css" integrity="sha384-fnmOCqbTlWIlj8LyTjo7mOUStjsKC4pOpQbqyi7RrhN7udi9RwhKkMHpvLbHG9Sr" crossorigin="anonymous">
```
lionleaf: Would it be better to host this locally? I prefer to avoid external resources, especially nice for offline work on airplanes (which I seem to do surprisingly frequently)

sigvef: Probably a good idea

lionleaf:
Thanks @sigvef :) I like the updated look of the comments as well.

However, I think we should keep hiding comments on dweets with many comments. But I agree with you in that having two paths of loading comments is unwieldy and a better solution would be to get all the comments with the dweet but hide some of them with js instead of loading new ones with AJAX

sigvef:
I updated this to still hide comments when there are many

sigvef:
Updated to fix all issues (except old-style visuals locally when submitting a new comment)

PR #445, opened 2019-09-29
Fix compressed remixes disabled button

kipkat:
Simple edit that allow people to remix with compressed code (doesn't disable the button when `#compressed` is in the first comment)

! **I HAVEN'T TESTED IF THIS WORKS, JUST AN IDEA** !


lionleaf:
This should now be fixed as part of #461 . Closing

PR #449, opened 2019-10-16
API #2

ben221199:
Removed the old API and fixed the favicon bug (missing / at beginning of relative url).


lionleaf:
Hi!

We've started the work on a new front-end in React. https://github.com/dwitter-net/dwitter-frontend 

And with that we are building a new API as well. It is already available at dwitter.net/apiv2beta/ .  It should contain everything needed for a full Dwitter client, and I've looked through this pull request and I think we provide all the features.

Let me know if you're still working on the Android client, if so give the new API a whirl and let us know if you need anything extra.

Closing this and the related pull requests.

PR #459, opened 2019-12-18
fix character count and remix button

KilledByAPixel:
count actual characters rather then string length

i think this should work but i don't have all the infrastructure to test it.

### DIFF
--- a/dwitter/static/js/feed.js
+++ b/dwitter/static/js/feed.js
@@ -3,8 +3,27 @@ function onDweetChanged() {
   var charCount = $(this).parent().parent().parent().find('.character-count')[0];
   var submitButton = $(this).parent().parent().find('.remix-button')[0];
 
-  charCount.textContent = this.value.length + '/140';
-  if (this.value.length > 140) {
+  let CountCharacters=string=>
+  {
+      let length = 0;
+      for (let i = 0; i < string.length;) 
+      {
+          let point = string.codePointAt(i);
+          let width = 1;
+          while (point)
+          {
+              ++width;
+              point >>>= 8;
+          }
+          i += width/2;
+          ++length;
+      }
+      return length;
+  }
+
+  let characterCount = CountCharacters(this.value);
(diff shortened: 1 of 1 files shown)


dwitter/static/js/feed.js
```diff
@@ -3,8 +3,27 @@ function onDweetChanged() {
...
+      for (let i = 0; i < string.length;) 
+      {
+          let point = string.codePointAt(i);
+          let width = 1;
+          while (point)
+          {
+              ++width;
+              point >>>= 8;
+          }
+          i += width/2;
```
lionleaf: Ok, so I'll admit this code wasn't immediately obvious to me; I needed to brush up on some UTF16 and spend some time to understand what was going on. 

Let me know if my understanding is right, and whether my suggestions useful. In any case I think we should add some comments here with pointers to UTF16 code points to help a future person reading through.

*My understanding: *
The magic here is that `codePointAt` will do the combination of surrogates for you, and the width counting is just trying to see if we got back a combined code point or not by checking if it is using more than 16 bits.

*Suggestions:*
I don't see why we need to count width in bytes instead of 16bit code points, so I believe something like the following could work:
```
if(point >>> 16){
     ++i; // This was a surrogate code point, skip next character
}
++length;
++i;
``` 

I also believe the current code will run into issues if you have a code point with a value < 256. As you'd be adding 0.5 to `i`. If we're keeping `width/2` we should add a rounding operation.

*Main suggestion:*
After some searching it looks like this code gives me the same results and is far simpler:  `[... string].length`

KilledByAPixel: wow, thats amazing! how did you discover that?

lionleaf: Came across it as a solution on some blog post about counting unicode characters. Can't claim ownership. But yeah, it seems like a much cleaner solution (with a good comment of why)

lionleaf:
Work continued in #461

PR #466, opened 2020-01-28
Make report buttons visible on all the themes

kipkat:
Now report buttons are readable in the new themes (the css wasn't updated for these new features).
I added a border to the buttons in the green theme (so they can be more visible).


lionleaf:
Deployed.

PR #476, opened 2020-05-10
Add automatic dark mode

veu:
Use _dark_ theme if no theme preference is set and dark mode is preferred (via [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)).

Features:
* Dark mode is used if it’s set in the OS.
* Preference from local storage overrides dark mode.
* "Next theme" button is aware of dark mode preference and will avoid switching to same theme.
* Theme handling code deals with adding and removing themes gracefully.
* Wrap everything in IIFE to avoid spamming global name space.

PS: I ran the linter but all I got were errors in files I didn’t touch.

### DIFF
--- a/dwitter/templates/base.html
+++ b/dwitter/templates/base.html
@@ -54,30 +54,31 @@
           theme = matchMedia('(prefers-color-scheme: dark)').matches ? themes[1] : themes[0];
         }
 
-        // Check for compatibility
-        if (window.localStorage) {
-          window.switchTheme = function() {
-            // Shift themes
-            theme = themes[(themes.indexOf(theme) + 1) % themes.length];
-            localStorage.setItem('theme', theme);
-            location.reload();
-            // Call this function in the li to change the mode
-          };
-        }
-
-        if (themeUrls[theme]) {
-          var l = document.createElement('link');
-          l.setAttribute('href', themeUrls[theme]);
-          l.setAttribute('type', 'text/css');
-          l.setAttribute('rel', 'stylesheet');
-          l.setAttribute('id', 'theme-style');
-          document.head.appendChild(l);
-        } else {
+        function updateTheme() {
           var l = document.getElementById('theme-style');
           if (l) {
             l.parentNode.removeChild(l);
           }
+          if (themeUrls[theme]) {
+            var l = document.createElement('link');
+            l.setAttribute('href', themeUrls[theme]);
+            l.setAttribute('type', 'text/css');
+            l.setAttribute('rel', 'stylesheet');
+            l.setAttribute('id', 'theme-style');
+            document.head.appendChild(l);
+          }
         }
+
+        updateTheme();
+
+        window.switchTheme = function() {
+          // Shift themes
+          theme = themes[(themes.indexOf(theme) + 1) % themes.length];
+          window.localStorage && localStorage.setItem('theme', theme);
+          updateTheme();
+
+          // Call this function in the li to change the mode
+        };
       })();
     </script>
     <div class="head-menu clear-fix">
--- a/dwitter/templates/base.html
+++ b/dwitter/templates/base.html
@@ -36,43 +36,41 @@
   </head>
   <body class="{% block body_class %}{% endblock %}">
     <script type="text/javascript">
-      // Check for compatibility
-      if (window.localStorage) {
-        var switchTheme = function() {
-          // Shift themes
-          var theme = localStorage.getItem('theme');
-          if (!theme) {
-            localStorage.setItem('theme', 'dark');
-          } else if (theme === 'dark') {
-            localStorage.setItem('theme', 'green');
-          } else if (theme === 'green') {
-            localStorage.setItem('theme', 'hcontrast');
-          } else {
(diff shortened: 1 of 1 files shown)


dwitter/templates/base.html
```diff
@@ -36,32 +36,38 @@
...
+          theme = matchMedia('(prefers-color-scheme: dark)').matches ? themes[1] : themes[0];
+        }
+
+        // Check for compatibility
+        if (window.localStorage) {
+          window.switchTheme = function() {
+            // Shift themes
+            theme = themes[(themes.indexOf(theme) + 1) % themes.length];
+            localStorage.setItem('theme', theme);
+            location.reload();
```
veu, line 63: While I’m at it, is there some compatibility issue with loading the CSS files dynamically without a reload? It wouldn’t be a big change to get rid of the reload otherwise.

stianjensen: It seems to me that the reload is just a quick way to re-run the code that sets the theme (which only runs once on boot). But that could easily be wrapped in its on function that is both run on boot and re-run when switching themes, I'd think.

I'll have a quick look in the change history of the file to see if there's any other reason I couldn't think of!

stianjensen: Yup, looks like it was just written like that from the start - I think that should be a quick change and just work :+1:

veu: Added it. Checked locally in FF, Chrome, and Safari.

dwitter/templates/base.html
```diff
@@ -36,43 +36,50 @@
...
+          }
         }
-      }
+
+        updateTheme();
+
+        window.switchTheme = function() {
+          // Shift themes
+          theme = themes[(themes.indexOf(theme) + 1) % themes.length];
+          window.localStorage && localStorage.setItem('theme', theme);
```
veu, line 77: I moved the check inline because the theme can now be changed without `localStorage`.

dwitter/templates/base.html
```diff
@@ -36,43 +36,50 @@
...
+          hcontrast: '{% static "theme_hcontrast.css" %}'
         };
+        // Use saved theme if localStorage is available
+        var theme = window.localStorage ? localStorage.getItem('theme') : null;
+        // Ignore unknown theme
+        if (themes.indexOf(theme) == -1) {
+          theme = null;
+        }
+        // Use dark mode preference if no theme is set
+        if (typeof theme != 'string' && window.matchMedia) {
```
lionleaf, line 53: Anyone who has looked at the other themes and gone back to the default will have `''` stored and won't be able to get the new automatic behavior. It might be worth being more explicit and having a 'light', 'dark' and 'automatic' theme and move anyone with `''` over to automatic? 

Could be done in a follow-up pull request

veu: As a one-time fix we could just use another key in local storage and ignore the old preference if it is `''` and that could easily be done in this PR. I wouldn’t add the _auto_ theme as long as it’s not displayed to the user because they’d cycle through the themes and see the “same” theme twice and wouldn’t know the difference.
 
As for long term behavior I’d prefer to move this to a place with a bit more room to show theme names and also explain what _automatic_ means.

lionleaf:
Good points. I like your suggestion of just ignoring "" and giving it a new
value as a one time fix for the pull request.

dwitter/templates/base.html
joeytwiddle: Yeah I agree that `''` can mean automatic, without creating a new `'automatic'` key.

But we do need a new `'light'` key.

joeytwiddle: If we don't want the user to see a duplicate theme, then it sounds like we must either:

- Not allow them to travel to `''` (automatic), or
- Not allow them to travel to `'light'` or `'dark'`, whichever is selected by automatic, or
- Provide some additional way to toggle the automatic theme selection, so if they are on `'light'` they can get back to `''`

dwitter/templates/base.html
joeytwiddle: There is no `location.reload()` now. Does it successfully overwrite all previously loaded themes with the latest selected theme? Great if so!

(For that to work, I think every theme needs to override all properties of the other themes. Perhaps that is the case now.)

dwitter/templates/base.html
veu: I’ll get rid of `''` for now. There won’t be an *automatic* theme right now, just a "user hasn’t selected a theme" which the users with `''` will fall back to. If they explicitly choose a theme, they won’t be able to go back for now but they’ll also know how to change it manually.
Yes, there is some more UI needed to do it properly and I’m not sure it should be fit into the dropdown menu.

dwitter/templates/base.html
veu: It doesn’t have to overwrite them because their CSS is removed first. In all browsers I tested, that worked, ie. they didn’t hold on to removed CSS rules.

veu:
I renamed the light theme. The existing preference will now be treated like any other unknown theme, ie. default to *automatic*.

PR #478, opened 2020-05-10
Split our dweet reply info to separate query

stianjensen:
Rather than joining on the dweet and user table one extra time in the
main query (which needs to be sorted and is already quite huge), this
splits out the query for loading information about the dweet to a
separate query. That means the total number of SQL queries on the
frontpage will increase by one, but I believe the total execution time
and amount of data to load from the database will be smaller, since the
data is joined in python instead.

### DIFF
--- a/dwitter/feed/views.py
+++ b/dwitter/feed/views.py
@@ -66,8 +66,7 @@ class DweetFeed(ListView):
         queryset = (
             queryset
             .select_related('author')
-            .select_related('reply_to')
-            .select_related('reply_to__author')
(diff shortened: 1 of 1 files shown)


stianjensen:
If you deploy this or #477 @lionleaf I'd like to not have them deployed too closely together, and know rough timestamps, so that I can compare perf data in New Relic and measure their effect.

lionleaf:
#477 deployed now, will hold off on deploying this for a while then so we can see the impacts of #477

stianjensen:
<pre align=center></pre>

Measurements from before and after deploy of this PR.

PR #480, opened 2020-05-10
Make dweet likes blank=True

stianjensen:
This allows you to edit a dweet in django-admin even though it has no
likes :'(


dwitter/models.py
```diff
@@ -32,7 +32,7 @@ class Dweet(models.Model):
     reply_to = models.ForeignKey("self", on_delete=models.DO_NOTHING,
                                  null=True, blank=True)
 
-    likes = models.ManyToManyField(User, related_name="liked")
+    likes = models.ManyToManyField(User, related_name="liked", blank=True)
```
lionleaf, line 35: Should it also be null=True while we're here?

stianjensen: Null doesn’t make sense for manytomany fields, since behind the scenes, thst isn’t actually a field ok this model, but backed by a seprste join table (where rows will only exist for actual non-null pairs).

So while most fields (except for textfields) need to be null=True when blank=True, that’s not the case for manytomany.

PR #481, opened 2020-05-11
Autoplay first dweet in the feed

lionleaf:
Also removes the call to newCode on load. While this is required to instrument the code, it already happens within `templates/dweet.html` so we were doing it twice.

This causes the first dweet to show up quicker, but since it starts before the page is fully loaded you might get some extra stutter.

I'd like feedback on whether that's worth it? Earlier dweet but extra stutter.


lionleaf:
Abandoning this for now. Not the right solution

PR #482, opened 2020-05-11
Only annotate num_likes on Top views

stianjensen:
The num_likes annotation is quite heavy, so avoiding it on pages where
it is not used seems smart.

With this, only the TopAllDweetFeed view should be particularly slow,
hopefully (the rest of the top views are fast since they start by
filtering out a lot of data based on posted at date before doing their
sorting).

### DIFF
--- a/dwitter/tests/feed/test_feed_views.py
+++ b/dwitter/tests/feed/test_feed_views.py
@@ -10,7 +10,7 @@ from django.utils import timezone
 from datetime import timedelta
 
 
-class DweetFeedTestCase():  # Not inheriting from TestCase, an abstract test class if you will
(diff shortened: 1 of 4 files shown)


stianjensen:
I think some tests will fail here, because they ensure that num_likes is always set. So I'll update those.

dwitter/tests/feed/test_feed_views.py
```diff
@@ -10,7 +10,7 @@
 from datetime import timedelta
 
 
-class DweetFeedTestCase():  # Not inheriting from TestCase, an abstract test class if you will
+class DweetFeedMixin:  # Not inheriting from TestCase, an abstract test class if you will
```
lionleaf, line 13: Yey, getting more django-esque 👍

PR #491, opened 2020-05-17
Change the remix-of field to embed the dweet instead of just pk

lionleaf:
Required for https://github.com/dwitter-net/dwitter-frontend/pull/33 

### DIFF
--- a/dwitter/serializers_v2.py
+++ b/dwitter/serializers_v2.py
@@ -29,12 +29,41 @@ class CommentSerializer(serializers.ModelSerializer):
         )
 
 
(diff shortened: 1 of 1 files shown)


sigvef:
You could make a new, smaller serializer (NestedDweetSerializer, SparseDweetSerializer, something like that), and embed that in the DweetSerializer to avoid recursion. That way, we can also only expose what we need in the nested version (e.g. perhaps not all of the comments).

lionleaf:
Good suggestion! See new changes. I included more than needed right away, as I can see the code etc of the original being useful for some views (say render it on hover) for either later additions or more importantly, other projects using the API.

PR #500, opened 2020-10-02
Big implementation of the first Dwitter Backend API (v1)

ben221199:
Conform the JSONAPI spec (https://jsonapi.org/).


lionleaf:
Thanks a lot for this contribution, and I'm aware this took significant effort. But I'm closing this for now, in favor of the new API we've built for the new frontend.

If you believe that's a mistake let's chat on Discord :) Happy to work together to make sure we get an API that has the features you might need for any other dwitter client!

ben221199:
I see. I sent you some information in on Discord.

PR #502, opened 2020-10-28
Makes it possible to log in with usernames over 30 characters

peterhgombos:
The maximum supported length of usernames in the database is 150 characters, however, the login screen does not currently support usernames over 30 characters. This makes it hard to test with longer usernames.

The registration form also caps the username that someone can enter to 30 characters, but this is only checked client side, so it is not impossible to get registered users with names that can't login by default.

### DIFF
--- a/dwitter/templates/registration/login.html
+++ b/dwitter/templates/registration/login.html
@@ -12,7 +12,7 @@
     </label>
     <input
         id="id_username"
-        maxlength="30"
+        maxlength="150"
         name="username"
         type="text"
         placeholder="..."


lionleaf:
Nice catch! 

As the login screen limits it to 30 characters, I would prefer to rather change the database to restrict it to 150 character and manually deal with any outliers. 30 characters seems more sane than 150 (that's more than a dweet!)

lionleaf:
Closing this, as we are deprecating this frontend in favor of https://github.com/dwitter-net/dwitter-frontend/

PR #504, opened 2020-11-30
fix: treat hashtags with a preceding : as anchors (fixes #474)

### LINKED ISSUE #474: Bug when links in comments contain anchors
Issue found by u/pavel
I recreated the bug here: [d/18324](https://www.dwitter.net/d/18324)
Description: when somebody wants to write a comment that contains a link like this:
```
http://www.website.com/page#anchor
```
#anchor will be detected as a hashtag and will break the automatic URL formatting. When posted, that url will turn into 
```
#anchor" rel="nofollow">http://www.website.com/page#anchor
```

Victim dweet: [d/18323](https://www.dwitter.net/d/18323)

This bug can be solved by adding the hashtag regex parsing before the url parsing or something like that

jellyedwards:
This should fix #474, it was a little tricky as hashtags#likethis and /u/even#likethis are allowed so to explain the regex I came up with:

```
(?:^|\s)    # go to the start of the line or the next space
(?=\S*#)    # lookahead to see is there a '#' after a bunch of non-whitespace characters?
(?!\S*:)    # lookahead to check aren't any ':' characters there (otherwise it's likely an anchor)
\S*         # skip any of the characters leading up to the '#'
# then do the hashtag grouping that was there before:
#(?P<hashtag>[_a-zA-Z][_a-zA-Z\d]*)
```

I've created a little test for it which passes along with the rest.

First time running a Django project & first time contributing so if I've messed up anything let me know!

- [X] Run `make lint` to ensure that all the files are formatted and using best practices.
- [X] Link to any issues this PR is solving.

### DIFF
--- a/dwitter/templatetags/insert_magic_links.py
+++ b/dwitter/templatetags/insert_magic_links.py
@@ -44,7 +44,8 @@ def insert_magic_links(text):
     )
     text = re.sub(
         r'(?P<text>'                                            # capture original pattern
-        r'#(?P<hashtag>[_a-zA-Z][_a-zA-Z\d]*)[^_a-zA-Z\d]?)',   # hashtag
+        # hashtag (that isn't preceded with a : so web anchors aren't considered)
+        r'(?:^|\s)(?=\S*#)(?!\S*:)\S*#(?P<hashtag>[_a-zA-Z][_a-zA-Z\d]*))',
         hashtag_to_link,
         text
     )
--- a/dwitter/templatetags/insert_magic_links.py
+++ b/dwitter/templatetags/insert_magic_links.py
@@ -44,7 +44,12 @@ def insert_magic_links(text):
     )
     text = re.sub(
         r'(?P<text>'                                            # capture original pattern
-        # hashtag (that isn't preceded with a : so web anchors aren't considered)
+        # (?:^|\s)    # go to the start of the line or the next space
+        # (?=\S*#)    # lookahead to see is there a '#' after a bunch of non-whitespace characters?
+        # (?!\S*:)    # lookahead to check aren't any ':' characters there (otherwise it's likely an anchor)
+        # \S*#        # skip any of the characters leading up to the '#'
+        # then do the hashtag grouping that was there before:
+        # (?P<hashtag>[_a-zA-Z][_a-zA-Z\d]*)        
         r'(?:^|\s)(?=\S*#)(?!\S*:)\S*#(?P<hashtag>[_a-zA-Z][_a-zA-Z\d]*))',
         hashtag_to_link,
         text
--- a/dwitter/tests/templatetags/test_insert_magic_links.py
+++ b/dwitter/tests/templatetags/test_insert_magic_links.py
@@ -192,9 +192,9 @@ class DweetTestCase(TestCase):
     def test_insert_magic_link_anchor_not_hashtag(self):
         self.assertEqual(
             '<a href="/h/start">#start</a> '
-            'https://www.site.com/page#anchor '
+            'https://www.example.com/page#anchor '
             '<a href="/h/end">#end</a>',
-            insert_magic_links('#start https://www.site.com/page#anchor #end')
+            insert_magic_links('#start https://www.example.com/page#anchor #end')
         )
 
     # mixed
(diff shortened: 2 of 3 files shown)


atesgoral:
Instead of making this URL-specific, could we instead say "the `#` needs to be preceded by a word boundary"?

dwitter/tests/templatetags/test_insert_magic_links.py
```diff
@@ -189,6 +189,14 @@ def test_insert_magic_single_character_hashtag(self):
             insert_magic_links('#1')
         )
 
+    def test_insert_magic_link_anchor_not_hashtag(self):
+        self.assertEqual(
+            '<a href="/h/start">#start</a> '
+            'https://www.site.com/page#anchor '
+            '<a href="/h/end">#end</a>',
+            insert_magic_links('#start https://www.site.com/page#anchor #end')
```
atesgoral, line 197: To be pedantic, https://tools.ietf.org/html/rfc6761:

```suggestion
            'https://www.example.com/page#anchor '
            '<a href="/h/end">#end</a>',
            insert_magic_links('#start https://www.example.com/page#anchor #end')
```

jellyedwards: Didn't know that, updated!

jellyedwards:
> Instead of making this URL-specific, could we instead say the `#` needs to be preceded by a word boundary (`\b`)?

That was my first attempt but some tests failed because `examplesLike#this` are allowed (where `#this` is the hashtag).

atesgoral:
> `examplesLike#this` are allowed.

Maybe this fix can be done as a breaking change. There isn't a reason (other than backwards compatibility) to support patterns like that, but the URL case is very common. So maybe the hashtag rule can be enforced as a "standalone #hash"?

jellyedwards:
I agree, it would be more consistent with other hashtag implementations, I just didn't want to break anything on my first contribution 😄

lionleaf:
I'm happy with a breaking change on hashtag behavior to align with more common practices.
I still think this is a good change to get in, it looks good to me.

Would you be OK with a rollout of this fix that only impacts new hashtags added and rendering of old hashtags? By that, I mean I'll leave the old links in the database. For example:
A comment with "http://example.com/site#anchor" would link that dweet with the #anchor hashtag. If I push this to the server with regenerating all the hashtag, said dweet will still be linked to the #anchor hashtag, show up if you go to h/anchor, but the #anchor link will be ignored as intended.

Let me know whether you'd want to move forward with this change as is, or attempt to simplify it and make hashtag parsing stricter.  (I could do some searches to see how many comments will be impacted by such a breaking change)

PS: Am I correct that your fix will still ignore example.com/something#anchor, i.e. without "https://"?
PPS: Apologies for being slow on this one, and thank you so much for the contribution!

dwitter/templatetags/insert_magic_links.py
```diff
@@ -44,7 +44,8 @@ def insert_magic_links(text):
     )
     text = re.sub(
         r'(?P<text>'                                            # capture original pattern
-        r'#(?P<hashtag>[_a-zA-Z][_a-zA-Z\d]*)[^_a-zA-Z\d]?)',   # hashtag
+        # hashtag (that isn't preceded with a : so web anchors aren't considered)
+        r'(?:^|\s)(?=\S*#)(?!\S*:)\S*#(?P<hashtag>[_a-zA-Z][_a-zA-Z\d]*))',
```
lionleaf, line 48: Would be worth adding the explanation from the PR as a comment here :)

jellyedwards:
Actually I don't think you need to make any DB changes, it should work with past and future comments.

Here's what it looks like with the original code:
image

.. and here's how it looks using this branch (no changing of DB entries): 
image

... so it should fix the existing ones!

dwitter/templatetags/insert_magic_links.py
jellyedwards: Done!

jellyedwards:
and yes it would ignore example.com/something#anchor

lionleaf:
> Here's what it looks like with the original code:
> image
> 
> .. and here's how it looks using this branch (same comment with no changing of DB entries):
> image

My point was simply that if you go to dwitter.net/h/Rela[...] that dweet will still be associated with that hashtag in the database. But I don't think that matters much.

This looks good. Thank you :)

lionleaf (approved):
Thanks a lot for your contribution

PR #512, opened 2021-05-30
Dweet perf improvements

ThomasBrierley:


ThomasBrierley:
Updated the first commit to remove `userFunction`'s unnecessary object.
Detailed explanations are included in each commit message.

sigvef:
Nice! Would be nice to test this on a set of dweets to see if there are any unforeseen side-effects from the change.

ThomasBrierley:
Agreed. I've so far checked for specific cases I was aware of using the following dweets:

For performance in Chrome and Firefox:
https://www.dwitter.net/d/22883

For 194b unicode unpacking:
https://www.dwitter.net/d/21705

For 185b unicode unpacking:
https://www.dwitter.net/d/12335

For unicode unpacking with surrounding code:
https://www.dwitter.net/d/22777

Regarding the new function definition, I can't forsee any behavioural difference beyond improved performance: `u` is still a function available to exploit as a function object or simply overwrite without breaking the loop.

Regarding the unicode unpacking, it only replaces substrings matching a `eval(unescape(escape`...`.replace(...)))` signature - additionally it evaluates the content of eval() to correctly accommodate all variants matching this signature.

Even so, more testing would be good, just not sure what to look for.

ThomasBrierley:
This should be my final addition. With these 3 improvements my test dweet https://www.dwitter.net/d/22883 runs flat out.
Firefox from 20 FPS -> 60 FPS
Chrome from 1 FPS -> 60 FPS
Some discussion on the dev channel in discord relating to this last change to loopbuster.

ThomasBrierley:
popped off the last commit so loopbuster perf can be tackled separately

ThomasBrierley:
And it's back again :smile: but simpler and faster.

lionleaf (approved):
As discussed on discord; looks good! :) 

Appreciate the contribution, thanks!

PR #513, opened 2021-05-31
Perf Compatibility Fixes

ThomasBrierley:
I've searched through all time top dweets all the way down to 24 awesomes' and only came across these two issues.

Variable compatibility - exploiting the locally scoped `code` variable:
https://www.dwitter.net/d/19823

Function compatibility - exploiting eval string to create a block scope `}{`:
https://www.dwitter.net/d/7493

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -154,6 +154,8 @@
           }
           displayError("");
           reset();
+          // for dweet compatibility
+          window.code = code;
       }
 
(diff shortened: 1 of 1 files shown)


sigvef (approved):
Nice job going through all those dweets! The changes look good.

ThomasBrierley:
Thanks. It's a nice smooth experience scrolling through them now.
Only thing that inhibits progress are latent ajax requests, but that can be fixed in the new front end.

PR #518, opened 2021-06-18
Dweet Perf Fix - Compressor compatibility

ThomasBrierley:
Compressor dweets e.g d/11852 rely on their compressed source string to
build their output. However The decompression perf improvement directly
mutates the "code" variable used by these compressors, breaking them.

This fix uses an independent local variable, preserving "code".


lionleaf (approved):
Nice fix

PR #519, opened 2021-09-07
Add register endpoint to apiv2 (beta)

lionleaf:
Enables the creation of new users through apiv2beta with POST requests to `users/`.
Also adds tests to verify some of the validation behavior.


dwitter/views_v2.py
```diff
@@ -17,10 +20,17 @@
...
 
 
-class UserViewSet(mixins.RetrieveModelMixin, viewsets.GenericViewSet):
+class UserViewSet(viewsets.ModelViewSet):
     queryset = User.objects.all()
     serializer_class = UserSerializer
 
+    def get_permissions(self):
+        # POST action to create new user need to be available to everyone
+        if self.request.method == 'POST':
```
sigvef, line 29: Does this also set AllowAny for detail actions like `set_email`? Perhaps we also need to check for `self.action == 'create'` as well here?

dwitter/views_v2.py
```diff
@@ -17,10 +20,17 @@
 from dwitter.utils import length_of_code
 
 
-class UserViewSet(mixins.RetrieveModelMixin, viewsets.GenericViewSet):
+class UserViewSet(viewsets.ModelViewSet):
```
sigvef, line 23: Does this accidentally expose update, partial_update, delete etc on user?

sigvef (changes requested):

dwitter/views_v2.py
lionleaf: Good catch. Been a while since I've properly read up on DRF; this would have been a very bad thing to merge 🙃

lionleaf: Added tests for this.

dwitter/views_v2.py
lionleaf: Good call; added some tests, and this was not an issue in practice since both `set_email` and `set_password` explicitly checks who is authorized. But adding the action check is still the right thing to do. Added.

dwitter/serializers_v2.py
```diff
@@ -8,11 +8,53 @@
...
+    def validate(self, data):
+        """
+        Checks to be sure that the received password and password2 are the same
+        And that the password meets auth.password_validation
+        """
+        errors = dict()
+        try:
+            # Note: 'password2' gets removed from the data here with pop
+            if data['password'] != data.pop('password2'):
+                errors['password'] = ('Passwords do not match')
```
sigvef, line 36: What's up with the parens here? If `errors['password']` should be a tuple of strings, then you need an extra comma: `('Passwords do not match',)`. If it should just be a string, we can remove the parens so that it doesn't look like an attempt at making up a tuple.

dwitter/serializers_v2.py
```diff
@@ -8,11 +8,53 @@
...
+            else:
+                validate_password(data['password'])
+        except ValidationError as e:
+            errors['password'] = list(e.messages)
+
+        if(data['username'] == '.' or data['username'] == '..'):
+            # avoids .. and . in url path
+            errors['username'] = ('Invalid username')
```
sigvef, line 44: Same here, with the tuples.

dwitter/tests/api/test_api2_user.py
```diff
@@ -0,0 +1,190 @@
...
+    def setUp(self):
+        self.client = Client()
+        self.user1 = User.objects.create_user(
+            username='user1',
+            email='user1@example.com',
+            password='user1Pass',
+        )
+
+    def get_user(self, user_id):
+        return self.client.get(APIV2_PATH + 'users/' + str(user_id) + '/')
```
sigvef, line 109: Since you asked for an in-depth review: this can also be written as
```python
f'{APIV2_PATH}users/{user_id}/'
```

sigvef (changes requested):

dwitter/tests/api/test_api2_user.py
lionleaf: Thanks, significantly easier to read ✅

PR #520, opened 2021-09-17
Add list of remixes to dweet apiv2 serializer

lionleaf:
Lists all remixes of a given dweet, only by ID.

This API change enables: https://github.com/dwitter-net/dwitter-frontend/pull/88 

### DIFF
--- a/dwitter/serializers_v2.py
+++ b/dwitter/serializers_v2.py
@@ -100,6 +100,7 @@ class DweetSerializer(serializers.ModelSerializer):
     id = serializers.ReadOnlyField(source='pk')
 
     remix_of = RemixOfSerializer(source='reply_to')
+    remixes = serializers.PrimaryKeyRelatedField(source='dweet_set', read_only=True, many=True)
 
     awesome_count = serializers.SerializerMethodField()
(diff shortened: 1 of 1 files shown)


dwitter/serializers_v2.py
```diff
@@ -100,6 +100,7 @@ class DweetSerializer(serializers.ModelSerializer):
     id = serializers.ReadOnlyField(source='pk')
 
     remix_of = RemixOfSerializer(source='reply_to')
+    remixes = serializers.PrimaryKeyRelatedField(source='dweet_set', read_only=True, many=True)
```
sigvef, line 103: We use this serializer to list dweets right? Perhaps we want to add prefetch_related for the dweet_set if we haven't already, for performance.

sigvef (changes requested):

dwitter/serializers_v2.py
lionleaf: The current setup doesn't ask for any of the dweets in this list unless a user explicitly clicks the remix list; which I expect is infrequent. Do you still think it's worth doing a prefetch for every request?

PR #524, opened 2021-10-29
Optimize apiv2 dweet requests

lionleaf:
Reduces the number of DB queries performed for `GET apiv2beta/dweets` from ~44 to 8

This PR does the following:
 - Adds `related_name=remixes` to the `reply_to` field in the Dweet model. This was done to make it more intuitive to read code that uses the reverse relation. 
   - (Note: When writing this I realize there's an inconsistency between the terminology of dweet "replies" and "remixes" in the code. Maybe we should rename the `reply_to` field `remix_of` later to make it even clearer?)
 - Removes the awesome_count field from the remix-linked dweet. This was not used and generated an extra query for each dweet.
 - Changed awesome_count and has_user_awesomed from serializer methods to fields generated by annotation. Previously these two functions added 3 queries for each dweet.
 - `prefetch_related` for `remixes`, the original N+1 problem that led me down this path.
 - Added a `prefetch_related` for `reply_to` including prefetch of the related author user profile. Before I did that I saw a lot of extra user queries

### DIFF
--- a/dwitter/views_v2.py
+++ b/dwitter/views_v2.py
@@ -4,6 +4,7 @@ from dateutil.parser import parse
 from django.contrib.auth.models import User
 
 from django.db.models import Prefetch, Count
+from django.db.models.expressions import Exists, OuterRef
 from django.utils import timezone
 
 from rest_framework import mixins, viewsets, status
@@ -70,10 +71,23 @@ class DweetViewSet(mixins.RetrieveModelMixin,
     queryset = Dweet.objects.all().select_related(
         'author',
     ).prefetch_related(
+        Prefetch('reply_to', queryset=Dweet.objects.select_related('author')),
         'likes',
+        Prefetch('remixes'),
         Prefetch('comments', queryset=Comment.objects.select_related('author'))
+    ).annotate(
+        awesome_count=Count('likes')
     ).order_by('-hotness')
+
     serializer_class = DweetSerializer
+
+    def retrieve(self, request, pk):
+        self.queryset = self.queryset.annotate(
+            has_user_awesomed=Exists(Dweet.objects.filter(
+                id=OuterRef('id'), likes__in=[request.user.id]))
+        )
+
+        return super().retrieve(request, pk)
 
     def create(self, request):
         code = request.data.get('code', '')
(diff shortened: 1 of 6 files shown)


dwitter/views_v2.py
```diff
@@ -70,18 +71,30 @@ class DweetViewSet(mixins.RetrieveModelMixin,
     queryset = Dweet.objects.all().select_related(
         'author',
     ).prefetch_related(
+        Prefetch('reply_to', queryset=Dweet.objects.select_related('author')),
```
sigvef, line 74: This one should probably be in select related instead of prefetch, since it is a forward fk.

```python
.select_related(
  ..,
  'reply_to',
  'reply_to__author'
)

File: dwitter/views_v2.py [hunk 2]
```diff
@@ -70,18 +71,30 @@ class DweetViewSet(mixins.RetrieveModelMixin,
...
         'likes',
```
sigvef, line 75: Maybe this one can be removed now that we annotate the count directly?

sigvef (changes requested):
Nice! Some comments.

dwitter/views_v2.py
lionleaf: Oh cool, I was looking for a way to do that :)

dwitter/views_v2.py
lionleaf: Did this comment land the wrong place?

Referring to `annotate(awesome_count=Count('likes'))`?  
I'll play around to see if it can get removed.

dwitter/views_v2.py
lionleaf: For sanity I checked:
Removing the prefetch goes from 8 to 19 queries.
Adding 'reply_to' to select_related brings it down to 13, and 'reply_to__author' further down to 7 (!)

dwitter/views_v2.py
lionleaf: Actually, were talking about the conditional annotate (line 131 in the old file)? In any case that one can be removed now :)

sigvef: I thought maybe we could remove fetching every single like, if we only used it for getting the count.

lionleaf: Oh yeah, you're right!
Down to 6 now:

1. django session
2. fetch user (from "rest_framework/views.py", in `perform_authentication`)
3. Subquery to fetch the dweets (3 joins)
4. Final dweet fetch (6 joins)
5. Fetch all the comments
6. Fetch all the remixes

sigvef (approved):

Issue #407, opened 2018-10-16
Draft?


ThomasBrierley:
I know the whole point is to keep it short and simple, but unlike human languages 140 chars or code can require quite a bit of hacking which takes more time and effort. A draft mode that saved before submission would be nice.

Just discovered Dwitter btw and I love the concept :smile: I am hooked

lionleaf:
Great idea. Definitely something that would be useful for a lot of people. You're talking about a full draft system here, right? You could have a "save as draft" that allows you to save the current dweet into a list of your drafts (even if it is more than 140 characters).

Another idea that's been brought up before is to just use localStorage to remember the current dweet you're working on so you can return later without losing progress. 

If anyone's interested, I think a draft system would be great! I might come around to implementing it some day, but I'm also happy to accept a pull request.

PS: Welcome to dwitter, and thanks for the suggestion!

dwitterer:
Might also be worth noting that there's always codegolf.tk if your dweets are too long for dwitter... And, of course, don't be afraid to post your code on our Discord server - more often than not, your code can be cut down into <140 characters if you get other people involved :P

PR #526, opened 2021-11-01
added: open links in new tab - accidentally navigating away from current page is a pita!

srmcgann:
added: open links in new tab - accidentally navigating away from current page is a pita!

### DIFF
--- a/dwitter/static/js/ajax-handling.js
+++ b/dwitter/static/js/ajax-handling.js
@@ -80,7 +80,7 @@ function getCommentHTML(comment) {
         '</div>' +
       '</button>' +
     '</form>' +
-    '<span class="comment-message">' + comment.urlized_text +
+    '<span class="comment-message">' + comment.urlized_text.replace('href=', 'target="_blank" href=') +
     '</span>' +
     '</li>';
 }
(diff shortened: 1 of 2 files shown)


lionleaf (approved):
I agree links in comments should open in a new tab! Thanks for the contribution :)

ThomasBrierley:
Is this referring to losing changes to the textarea upon navigation?

Interestingly Firefox has saved me from this a number of times because it autofills previously changed fields based on history navigation... for a short time at least.

I'm wondering if we should implement something similar as a more general solution to this type of problem for all browsers.
I'm imagining some combination of `localStorage` with `history` and dweet id to differentiate between windows into the same dweet... although that would not uniquely identify it since the only history property we can use is `history.length` i think. Maybe with a timestamp to limit how long until it's "forgotten" would minimise mix-ups, i.e so we try to only affect accidental navigation of any kind... assuming people hit back, do they?

Just an idea. I'm aware of feature creep, so we shouldn't implement it if it doesn't sound useful enough.

lionleaf:
You could also lose your scroll location etc.

But I think a localstorage draft system could make sense. This is the closest issue I could find about it: https://github.com/lionleaf/dwitter/issues/407

PR #530, opened 2021-11-15
Add deletion of dweets and comments to apiv2

lionleaf:
- [x] Add tests

Adds comment and dweet deletion to apiv2. Required for https://github.com/dwitter-net/dwitter-frontend/pull/112

### DIFF
--- a/dwitter/serializers_v2.py
+++ b/dwitter/serializers_v2.py
@@ -66,6 +66,7 @@ class UserSerializer(serializers.ModelSerializer):
         user = self.context['request'].user
         if user.is_authenticated and user.id == instance.id:
             value['email'] = user.email
+            value['is_staff'] = user.is_staff
         return value
 
 
@@ -128,11 +129,20 @@ class RemixOfSerializer(serializers.ModelSerializer):
         model = Dweet
         fields = (
             'author',
+            'deleted',
             'code',
             'id',
             'posted',
             'remix_of',
         )
+
+    def to_representation(self, instance):
+        value = super().to_representation(instance)
+        if instance.deleted:
+            value['code'] = ''
+            value['comments'] = []
+
+        return value
 
 
 class DweetSerializer(serializers.ModelSerializer):
@@ -150,6 +160,7 @@ class DweetSerializer(serializers.ModelSerializer):
         model = Dweet
         fields = (
             'author',
+            'deleted',
             'awesome_count',
             'code',
             'comments',
@@ -159,3 +170,12 @@ class DweetSerializer(serializers.ModelSerializer):
             'remixes',
             'has_user_awesomed',
         )
+
+    # If we implement actual database deletion of dweets,
+    # we can remove this
+    def to_representation(self, instance):
+        value = super().to_representation(instance)
+        if instance.deleted:
+            value['code'] = ''
+            value['comments'] = []
+        return value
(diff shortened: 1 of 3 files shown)


dwitter/serializers_v2.py
```diff
@@ -66,6 +66,8 @@ def to_representation(self, instance):
         user = self.context['request'].user
         if user.is_authenticated and user.id == instance.id:
             value['email'] = user.email
+            value['is_staff'] = user.is_staff
+        # 'email' is set to write-only so won't be included for other users
```
sigvef, line 70: Not sure if I understand this comment?

dwitter/views_v2.py
```diff
@@ -82,7 +83,17 @@ class DweetViewSet(mixins.RetrieveModelMixin,
     serializer_class = DweetSerializer
 
     def retrieve(self, request, pk):
-        self.queryset = self.queryset.annotate(
+        # For single dweet view, also include deleted dweets
```
sigvef, line 86: Why? 🤔

dwitter/views_v2.py
```diff
@@ -150,10 +161,24 @@ def list(self, request):
...
+    def destroy(self, request, *args, **kwargs):
+        if not request.user.is_authenticated:
+            raise PermissionDenied()
+
+        dweet = self.get_object()
+
+        # Only allow deletion of own dweets
+        # unless the deleter is a mod
+        if request.user == dweet.author or request.user.is_staff:
+            dweet.delete()
```
sigvef, line 173: .delete() on an instance is built-in functionality from django I think. Will this actually delete from the db instead of setting the deleted flag to True?

sigvef (changes requested):
Some questions

dwitter/views_v2.py
lionleaf: For two reasons:
1. Feature parity with previous frontend, which shows you a [deleted dweet] page rather than a 404 for deleted dweets
2. To allow you to follow remix chains through deleted dweets.

I'm happy to have the discussion of whether we should permanently delete all records of the dweets instead,  but that would break remix chains .

dwitter/views_v2.py
lionleaf: Soft delete was implemented by overriding the delete() function. 
I've tested the new [deleted] views locally, and I can't see how that would work if this was a full delete.
https://github.com/lionleaf/dwitter/blob/a70a82c4854dd192dd9e910df515d1082690e85e/dwitter/models.py#L50

dwitter/serializers_v2.py
lionleaf: When looking at that code I was suddenly confused about why email wasn't just returned to all users, and left that comment. I don't think the comment belongs there so it's been removed.

We also have a test now that will break if we start leaking email addresses :)

dwitter/views_v2.py
```diff
@@ -199,16 +223,31 @@ def report(self, request, pk=None):
...
+    def destroy(self, request, *args, **kwargs):
+        if not request.user.is_authenticated:
+            raise PermissionDenied()
+
+        comment = self.get_object()
+
+        # Only allow deletion of own dweets
+        # unless the deleter is a mod
+        if request.user == comment.author or request.user.is_staff:
+            comment.delete()
```
sigvef, line 242: Hm, I guess we don't have soft delete for comments iirc. Should we perhaps have that too?

sigvef (changes requested):

sigvef (approved):

PR #533, opened 2021-11-23
New API endpoint for user stats

rep-movsd:
Implement an endpoint  `/stats?username=foo` that returns total dweet count and total awesome count for that user

### DIFF
--- a/dwitter/urls.py
+++ b/dwitter/urls.py
@@ -54,6 +54,7 @@ urlpatterns = [
     url(r'^', include('dwitter.feed.urls')),
     url(r'^api/', include(router.urls)),
     url(r'^apiv2beta/api-token-auth/', authtoken_views.obtain_auth_token),
+    url(r'^apiv2beta/stats/', views_v2.StatsAPI.as_view(), name='stats'),
     url(r'^apiv2beta/', include(router_v2.urls)),
 ]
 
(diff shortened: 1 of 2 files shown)


lionleaf (changes requested):
Thanks! This looks good and runs locally :) 

However,  I would ask for one change to the URL path to stay consistent with the way the rest of the apiv2 does user lookup, could we change it to `apiv2beta/users/{username}/stats`? 

This can be done in one of two ways:
1. Make stats an @action method of the `UserViewSet` class and remove the custom `url()`. Something like this (coded in browser):
```
@action(detail=True, methods=['POST'])
    def stats(self, request, pk=None):
        user = self.get_object()
        queryset = Dweet.objects.all()
        [... rest of your function]
```
2. Simply change the URL pattern to `apiv2beta/(?P<url_username>[\w.@+-]+)/stats`  and then use the url_username parameter.

rep-movsd:
Changes applied as requested

lionleaf (approved):
Thanks!

lionleaf:
You're failing the lint check  (run `make lint` locally, or check thee Travis build directly: https://app.travis-ci.com/github/lionleaf/dwitter/builds/242320839#L329 )

Ready to merge merge and deploy once lint and tests are passing.

PR #534, opened 2021-12-02
Start dweet paused and hidden if URL parameter "spoiler" is specified

### LINKED ISSUE #448: A #spoiler suggestion
### `#spoiler` 
**will play the dweet only when clicked on canvas for every dweet tagged with #spoiler**
the autoplay will be disabled & `recieveMessage` should work different. Is it too hard to implement this? _Similar to the discordapp's spoiler system. This is useful for #394 too_

A image should be shown before clicking the canvas. Something like:
image

I think it is useful for heavy dweets, audio dweets, nsfw jokes dweets & every other kind of dweet that shouldn't be played automatically. What do you guys think?

rep-movsd:
Fixes #448 for the backend

Adding spoiler=1 to the dweet url will keep it paused and hidden until user clicks

### DIFF
--- a/dwitter/templates/dweet/dweet.html
+++ b/dwitter/templates/dweet/dweet.html
@@ -8,13 +8,33 @@
         outline:0;
         overflow:hidden;
       }
+
       canvas {
         width: 100%;
         background: white;
       }
+
+      #curtain {
+        width: 100%;
+        height: 100%;
+        background-color: black;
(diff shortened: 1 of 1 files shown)


lionleaf (approved):
Thanks! 

You forgot to set the overlay to hidden for non-spoiler dweets, but happy to merge and just fix that before deploying :)

PR #535, opened 2021-12-04
Fix recording dweets on Chrome

diceroll123:
Currently when recording is done processing, you'll find this in the console: `Download is disallowed. The frame initiating or instantiating the download is sandboxed, but the flag ‘allow-downloads’ is not set. See https://www.chromestatus.com/feature/5706745674465280 for more details.`

---


lionleaf (approved):
Thanks for this fix!

lionleaf:
deployed :)

